PR #7, opened 2015-07-16
Fix for import svg image leading to unattached dragtool

ibrierley:
If you import an svg image, it loads it, but then the dragtool appears on its own, not connected to the newly imported element.

When the element is imported, it isn't known 'what' this element is, so the change is to make the method return the element, so we can then use it, to 'select it, once the load is complete.

I cannot see any other methods reliant on the return value of this (which previously returned true rather than the element).

### DIFF
--- a/editor/svgcanvas.js
+++ b/editor/svgcanvas.js
@@ -5083,7 +5083,8 @@ this.importSvgString = function(xmlString) {
 		return false;
 	}
 
-	return true;
+	// we want to return the element so we can automatically select it
+	return use_el;
 };
 
 // TODO(codedread): Move all layer/context functions in draw.js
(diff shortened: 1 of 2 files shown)


editor/svgcanvas.js
```diff
@@ -5083,7 +5083,8 @@ this.importSvgString = function(xmlString) {
 		return false;
```
codedread: Please change this to return null

editor/svgcanvas.js
```diff
@@ -5083,7 +5083,8 @@ this.importSvgString = function(xmlString) {
...
 	}
 
-	return true;
+	// we want to return the element so we can automatically select it
```
codedread: Please update the docs for this function (which still say that it returns true / false)

codedread:
Sorry for the delay on this!

PR #13, opened 2015-09-24
Added class support

straps:
Classes are very useful when you need to group items and use them programmatically.

Merge this please, it just works, thanks

### DIFF
--- a/editor/svg-editor.js
+++ b/editor/svg-editor.js
@@ -2992,7 +2992,7 @@ TODOS
 					return false;
 				}
 
-				if (attr !== 'id' && attr!=='class') {
+				if (attr !== 'id' && attr !== 'class') {
 					if (isNaN(val)) {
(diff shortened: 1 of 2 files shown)


codedread:
Do the toolbar components wrap properly?

editor/svg-editor.js
```diff
@@ -2991,7 +2992,7 @@ TODOS
 					return false;
 				}
 
-				if (attr !== 'id') {
+				if (attr !== 'id' && attr!=='class') {
```
codedread: nit: can you please space things out here like the code around it:

  attr !== 'class'

straps:
What do you mean? HTML structure is the same as ID field...

editor/svg-editor.js
straps: done, commit ce009a4

PR #16, opened 2015-10-13
Explain that some functionality is still used on Google Code

MrRio:
brettz9 explained the situation with Google Code, have updated the readme and also switch it over to Markdown.


codedread:
Btw, I have no preference about where issues are tracked - seems like migrating fully off of Google Code would be a good idea though

Issue #29, opened 2015-11-30
Icons are broken on Safari


progers:
In the 2.8 release, all icons are broken in Safari due to a re-definition of xmlns. I bisected this bug down to https://github.com/SVG-Edit/svgedit/commit/b1f88d2b947f177688b1725326e7849280ea9a0b.

PR #30, opened 2015-11-30
Fix icons in Safari due to a redefinition of the xmlns attribute

progers:
https://github.com/SVG-Edit/svgedit/commit/b1f88d2b947f177688b1725326e7849280ea9a0b switched to using XMLSerializer which will create two xmlns attributes if a viewbox attribute is added with a non-null namespace. Two xmlns attributes results in a parse error in Safari which breaks all icons.

This change simply switches the viewBox setAttributeNS call to use a null namespace, per the xml-names11 recommendation. Without a rewrite, unit testing this change in jquery-svgicons is not really possible so I've just added a comment about why a null namespace is used. I've tested this change manually on firefox, safari, and chrome.

This change also removes the deadcode svgcontent variable that https://github.com/SVG-Edit/svgedit/commit/b1f88d2b947f177688b1725326e7849280ea9a0b left around.

This pull request fixes https://github.com/SVG-Edit/svgedit/issues/29

### DIFF
--- a/editor/svgicons/jquery.svgicons.js
+++ b/editor/svgicons/jquery.svgicons.js
@@ -290,7 +290,9 @@ $(function() {
 					holder = $('#' + id);
 					var svg = elem.getElementsByTagNameNS(svgns, 'svg')[0];
 					var svgroot = document.createElementNS(svgns, "svg");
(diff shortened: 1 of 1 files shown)


progers:
@brettz9, can you review this and click "merge" if it looks good to you?

PR #32, opened 2015-11-30
Update lang.fr.js

StalderT:
fix "delete" "move_down" "move_back" french translation


progers:
This looks good. I don't speak french but google translate says this is some belles français :)

I think we'll re-release svg-edit 2.8 as 2.8.1 soon due to a safari bug with the icons so lets get this merged.

PR #34, opened 2015-12-01
Intersection list

codedread:
1) use getIntersectionList() when available (Opera, Chrome, Safari).
2) remove all suspendRedraws() - that API is removed in SVG 2

### DIFF
--- a/editor/select.js
+++ b/editor/select.js
@@ -93,6 +93,7 @@ svgedit.select.Selector.prototype.reset = function(e) {
 	this.selectorGroup.setAttribute('display', 'inline');
 };
 
+
 // Function: svgedit.select.Selector.updateGripCursors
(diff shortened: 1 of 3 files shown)


progers:
This change looks great, just one minor comment nit.

Is this covered by a unittest? If so--does the test pass in both firefox and chrome?

codedread:
This PR fixes Issue #36

codedread:
"Is this covered by a unittest?" 

The sad answer is no - we never really wrote unit tests back in the early days of SVG-edit.  I only started when I began trying to disentangle all the code and split things out into modules.  I'd prefer to hold off on writing unit tests for svgcanvas since it's such a tangled mess still...

progers:
And we're in!

Issue #31, opened 2015-11-30
Switch svg-edit to work over https


progers:
The currently svg-edit release is served off http:
http://svg-edit.github.io/svgedit/releases/svg-edit-2.8/svg-editor.html

We should switch this to https but two image library urls need to be switched to https in editor/extensions/ext-imagelib.js. I verified both of these urls can be served over https.

Once this is fixed, README.md can be updated to point to the https version of svg-edit.

PR #38, opened 2015-12-02
Switch imagelib to https urls, update project to be hosted over https

progers:
This pull request fixes issue https://github.com/SVG-Edit/svgedit/issues/31 so that svgedit can be served securely. There are side-channel attacks when using protocol relative urls (http://www.paulirish.com/2010/the-protocol-relative-url/) so this patch goes whole-hog on https.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 ===
 SVG-edit is a fast, web-based, javascript-driven SVG drawing editor that works in any modern browser.
 
-### [Try SVG-edit here](http://svg-edit.github.io/svgedit/releases/svg-edit-2.8/svg-editor.html)
+### [Try SVG-edit here](https://svg-edit.github.io/svgedit/releases/svg-edit-2.8/svg-editor.html)
 
 (Also available as a [download](https://github.com/SVG-Edit/svgedit/releases/download/svg-edit-2.8/svg-edit-2.8.zip) in [releases](https://github.com/SVG-Edit/svgedit/releases)).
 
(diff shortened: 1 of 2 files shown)


adrianbj:
Looks like the link to the https version of the demo has been reinstated, but the code the editor is running off is the 2.8 release, rather than the latest master, so the problem of the image libraries not working is back.

progers:
I'm working on a 2.8.1 release right now, should be up shortly.

2.8 introduced a bug that broke all icons in Safari (https://github.com/SVG-Edit/svgedit/issues/29) so we needed to do a re-release anyway, and I figured this would be nice to have.

adrianbj:
Sorry, didn't mean to hassle - just noticed and thought I should flag in case it had been overlooked.

Thanks for all your recent efforts on svgedit. I really should look back at some of the modifications I made to the version I am running and contribute these at some point - some things related to breaking apart imported svg <symbol> elements. Anyway getting OT now so will leave it there :)

progers:
Oh, no hassle at all :)

I think this is all ready now. Could you please give it a try and make sure it works for you too?

adrianbj:
The open clipart library seems to be mostly fine although the links to the SVGs themselves are giving Mixed Content warning.

The IAN one doesn't actually load any SVGs at all - looks like it's an issue with getting jquery loaded as it's requesting that insecurely. I'll modify the links and see if I can get things working.

adrianbj:
Ok, the IAN Library is fixed working again now! 
Now I just need to put some time aside to improve the horribly dated interface :)

Issue #26, opened 2015-11-29
Chrome is unusable when selecting lots of paths


codedread:
Open SVG-edit with the SVG file pasted below and then drag-select a bunch of the paths.  You can see that Chrome has a hard time rendering the select boxes and keeping up with the mouse.  No similar problem exists in Firefox which keeps up just fine.
# 

<svg width="500" height="500" xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg">
 
 <g stroke="null">
  <title>Features</title>
  <path stroke="#000000" fill="none" stroke-width="2" d="m148.72322,202.77881c3.67596,-7.00858 4.05699,-8.48969 7.73283,-15.49817c4.82916,-5.13416 10.29091,15.0087 12.77013,22.47559" id="svg_18"/>
  <path stroke="#000000" fill="none" stroke-width="2" d="m134.34709,188.39349c2.60612,-9.00548 4.22293,-15.91933 9.37553,-22.53543c5.15273,-6.61609 7.35718,15.71062 10.2251,23.18295" id="svg_7"/>
  <path stroke="#000000" fill="none" stroke-width="2" d="m136.16826,164.55119c1.1557,-5.18611 3.70772,-8.86615 7.38377,-15.87468c4.82896,-5.13423 10.29083,15.00858 12.77013,22.47551" id="svg_9"/>
  <path stroke="#000000" id="svg_5" fill="none" stroke-width="2" d="m130.94594,204.61511c3.67603,-7.00833 9.60486,-25.04381 14.08055,-16.27576c4.47578,8.76801 2.61897,14.25388 6.4225,23.25299"/>
  <path stroke="#000000" fill="none" stroke-width="2" d="m152.97641,175.37897c3.67586,-7.00858 4.05688,-8.48958 7.73283,-15.49817c4.82898,-5.13432 10.29091,15.00867 12.77014,22.47546" id="svg_8"/>
  <path stroke="#000000" fill="none" stroke-width="2" d="m150.13936,156.1627c3.67606,-7.00853 4.05699,-8.48958 7.73294,-15.49815c4.82898,-5.13428 10.29082,15.00871 12.77002,22.47554" id="svg_10"/>
[...]

PR #46, opened 2015-12-07
Only disable overview panel in Chrome 48-, not all Chrome

codedread:
Better fix for issue #26 now that Chrome has a fix scheduled for release.


progers:
Nice! Thanks for the quick fix

Issue #59, opened 2016-01-20
Undo Redo Key Binding


GoTeamScotch:
Issue: The keypresses "Z" and "Y" trigger undo/redo respectively BUT Z is also tied to Zoom, so while pressing Z you get both actions (undo & zoom) triggered. So, when you go to hit the zoom key, your last action is undone! 

I fixed this issue by changing this:

```
{sel:'#tool_undo', fn: clickUndo, evt: 'click', key: ['Z', true]},
{sel:'#tool_redo', fn: clickRedo, evt: 'click', key: ['Y', true]},
```

To

```
{sel:'#tool_undo', fn: clickUndo, evt: 'click'},
{sel:'#tool_redo', fn: clickRedo, evt: 'click'},
```

Found in svg-editor.js around line 4020 (when not using compiled assets).

Undo / Redo can still be triggered by pressing [modKey] + Z or [modKey] + Y respectively so you're not losing anything with this change.

codedread:
Can you send a pull request?

GoTeamScotch:
Done.

brettz9:
Looks like this can be safely closed with the PR having been merged... Feel free to comment if otherwise...

PR #60, opened 2016-01-20
#59 Remove double-bind of keypresses Z and Y

GoTeamScotch:
Patch for Issue #59. https://github.com/SVG-Edit/svgedit/issues/59


GoTeamScotch:
Was the compiled script also edited? I did not do that in my pull request. I always use the separated/non-minified assets in my development site. I'm not sure the exact process of adding scripts together and then minifying them (their order, what should be included and what shouldn't).

GoTeamScotch:
@codedread - did you see my last message? This commit will not affect most users.

codedread:
We do releases periodically, see the build process, I believe it's documented in the code.  That's when compiled scripts would be updated.

PR #69, opened 2016-02-22
Remove fix that cloned path segments and add tests to show it works

### LINKED ISSUE #66: Path Editing broken
1) Make a polygon using the path tool
2) Double-click the path so the individual points are selected on the path
3) Try to move a node in the path.

Observed:  The path node tries to move but jitters back to its starting point.  If you double-click the point to turn the segment into a curve, then the path node is movable.  This is a horrible workaround.

progers:
This largely reverts fbbd740c794e27a384eea0abd7987a1515b0a6b4 and adds tests (passing in firefox, chrome, and safari) to show svgedit still works. I also manually tested the path tool in each browser and didn't notice any bugs.

I am not sure which bug fbbd740c794e27a384eea0abd7987a1515b0a6b4 was fixing but if it crops up again we now have the test infrastructure to better test it.

This fixes #66.

### DIFF
--- a/editor/path.js
+++ b/editor/path.js
@@ -478,16 +478,12 @@ svgedit.path.Segment.prototype.update = function(full) {
 svgedit.path.Segment.prototype.move = function(dx, dy) {
 	var cur_pts, item = this.item;
 
-	// fix for path tool dom breakage, amending item does bad things now, so we take a copy and use that. Can probably improve to just take a shallow copy of object
-	var cloneItem = $.extend({}, item);
-
-        if (this.ctrlpts) {
-                cur_pts = [cloneItem.x += dx, cloneItem.y += dy,
-                        cloneItem.x1, cloneItem.y1, cloneItem.x2 += dx, cloneItem.y2 += dy];
-        } else {
-                cur_pts = [cloneItem.x += dx, cloneItem.y += dy];
-        }
-	//fix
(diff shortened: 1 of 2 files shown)


ibrierley:
I think that was one of my workarounds, oddly I can't see the bug in github issues (maybe it was referenced from Google group, but I could have sworn there was an issue on git). The bug was where after about 3 points created and adjusted in the pathtool, there would be an error in the console, and the path would be deleted. The cloning was a way to get around a subtle issue under the hood that I found hard to understand iirc, so was never quite comfortable with my fix, so glad its all working correctly now.

progers:
Thanks for confirming ibrierley. That case seems to work with the patch but I think there's a good chance a real bug is hiding down there somewhere. It's also possible that it was a browser bug that's been fixed.

I think the best approach will be to land this patch, and then separately look into any bugs that crop up as a result.

ibrierley:
Is there a version running with the commit out of interest, to take a peek ?

progers:
@ibrierley, sure thing:
https://pr.gg/svgedittest/svg-editor.html

ibrierley:
Thanks, can confirm that the browsers I had been testing in all work with
this fine.

magnebra:
It works fine for me too.

progers:
In we go! Thank you both.

PR #71, opened 2016-02-25
Ellipse preserve 0 rx ry

### LINKED ISSUE #70: Ellipses with rx/ry = 0 is invalid after cleanupElement
Ellipses require the rx/ry attribute and the "0" value has a specific meaning (do not display). Removing these attributes in cleanupElement results in invalid svg.
[https://www.w3.org/TR/SVG/svgdtd.html](svg DTD)
[https://www.w3.org/TR/SVG/shapes.html#EllipseElement](svg doc)

magnebra:
This fixes #70

### DIFF
--- a/editor/svgutils.js
+++ b/editor/svgutils.js
@@ -630,8 +630,8 @@ svgedit.utilities.cleanupElement = function(element) {
 		'stroke-linecap':'butt',
 		'stroke-opacity':1,
 		'stroke-width':1,
-		'rx': 0,
-		'ry': 0
+		'rx':0,
+		'ry':0
 	};
 
 	var nodeName = element.nodeName ? element.nodeName : '';
--- a/editor/svgutils.js
+++ b/editor/svgutils.js
@@ -620,7 +620,6 @@ svgedit.utilities.assignAttributes = function(node, attrs, suspendLength, unitCh
 // Parameters:
 // element - DOM element to clean up
 svgedit.utilities.cleanupElement = function(element) {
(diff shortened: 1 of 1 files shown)


editor/svgutils.js
```diff
@@ -634,6 +634,13 @@ svgedit.utilities.cleanupElement = function(element) {
 		'ry':0
 	};
 
+	var nodeName = element.nodeName ? element.nodeName : '';
+	if (nodeName == 'ellipse') {
```
progers: (Nit) this should do the trick without creating a local string:
if (element.nodeName === 'ellipse') {

editor/svgutils.js
```diff
@@ -634,6 +634,13 @@ svgedit.utilities.cleanupElement = function(element) {
...
+		// Ellipse elements requires rx and ry attributes
```
progers: I'm afraid I don't understand what this PR is doing. cleanupElement removes the rx and ry attributes if they are set to zero. This PR changes cleanupElement to leave rx and ry as zero--won't the rendering still be disabled since the default rx and ry values are zero?

Can you describe how to reproduce the bug that's being fixed?

magnebra:
The svg definition requires the rx and ry attributes to be present on the ellipse element, so removing them will create a svg document that is invalid.
I found the problem when I passed a svg drawing from svgedit through the batik library for processing (which just crashes in this case).

progers:
I see now, thanks for the explanation.

PR #85, opened 2016-03-11
Change the text input bindings to catch more input events

### LINKED ISSUE #84: Using alt+keycode on windows does not update text
When creating a text object, using ALT+keycode for input does not update the text, one must press another key before the ALT-keycode is shown.

magnebra:
fixes #84 by binding more events to the text input

### DIFF
--- a/editor/svg-editor.js
+++ b/editor/svg-editor.js
@@ -2961,7 +2961,7 @@ TODOS
 				svgCanvas.setSegType($(this).val());
 			});
 
-			$('#text').keyup(function() {
+			$('#text').bind("change keyup input", function() {
 				svgCanvas.setTextContent(this.value);
 			});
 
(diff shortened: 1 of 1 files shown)


magnebra:
In my test, "change" was not enough to catch ALT+keycode

progers:
I haven't tested this, but various stackoverflow examples show alt+keycode combinations working with a single event (e.g., http://stackoverflow.com/questions/11412044/javascript-alt-key). Could an approach like that work?

I believe that capturing all 3 events will fix this alt bug, but it'll also cause duplicate work in the common case.

magnebra:
I removed the 'change' binding as this was not required.
The example you linked seem to handle things like ALT+M, but my problem was when using ALT+numpad to enter characters by its keycode. For example ALT+0176 on the numpad will insert a degree character (this is a windows thing as far as I know).

PR #90, opened 2016-03-14
Check for deleted defs after paste, fixes #82

### LINKED ISSUE #82: Pasting a line with markers produce invalid svg if original line was deleted
To reproduce:
- Create a line with markers
- Copy the line
- Delete the line
- Save the drawing
- Paste the line
The marker defs is now missing from the svg.

My guess is that the defs are not copied along with the line, so when pasting the line, it refers to a defs that were removed when saving the drawing.

Possible solutions would be:
- copy the defs along with the element
- don't delete defs for elements on the clipboard

magnebra:


progers:
This pull request looks good but can you please add a better description? Something that describes why restoring the defs is the correct thing to do in this case.

magnebra:
When pasting an element, we must make sure any defs which are referenced by the element is actually present. In some cases the defs might have been deleted (as described in #82) so we call restoreRefElems to restore any deleted defs referenced by the element.

progers:
Thanks!

Issue #94, opened 2016-03-21
Support for custom namespaces


iuyiuy:
I have a need for having some elements with some added meaning, that I would like to manage myself. This would normally be done with namespaces, but there doesn't seem to be a way to add namespaces and elements. It's possible to add to the svgedit.NS object from the config file, but this won't allow the namespace declaration in the svg tag, nor the actual elements using this namespace.

I could always manage the custom namespaces, either in extensions or from the output afterwards.

praneybehl:
I support that :+1:

flintobrien:
We use an extension to add a custom namespace (see example below). There is one bug that prevents this from working. See [pull request #110](https://github.com/SVG-Edit/svgedit/pull/110).

#### Example Extension:
```
svgedit.editor.addExtension( "ext-fisheye-base", function ( S ) {
   svgedit.sanitize.addWhiteListException("g", 'http://www.totalgrid.org', 'some-attribute');
   svgedit.sanitize.addWhiteListException("g", 'http://www.totalgrid.org', 'another-attribute');
   svgedit.sanitize.addWhiteListException("rect", 'http://www.totalgrid.org', 'some-attribute');
   svgedit.sanitize.addWhiteListException("rect", 'http://www.totalgrid.org', 'another-attribute');
   ...
}
```

flintobrien:
I just realized that [sanitize.addWhiteListException()](https://github.com/gec/svgedit/blob/fisheye/editor/sanitize.js#L105) is only in the [GEC fork, branch fisheye](https://github.com/gec/svgedit/tree/fisheye) of svgedit. I can do a pull request if others think it would be helpful as part of a more complete solution.

iuyiuy:
I implemented this in pull request no. 97:
https://github.com/SVG-Edit/svgedit/pull/97

flintobrien:
The ```svgedit.ignoredNS``` solution in [#97](https://github.com/SVG-Edit/svgedit/pull/97) is interesting. What are people's thoughts for the alternative solution of extending the whitelist either as a callable API or config?

Here's an example config:
```
svgEditor.setConfig({
  extendedNamespaces: {
    MYNS: 'http://mynamespace.com'
  },
  extendedWhiteList: {
    "g": ["MYNS:some-attribute", "MYNS:another-attribute"],
    "rect": ["MYNS:some-attribute", "MYNS:another-attribute"]
  }
});
```

iuyiuy:
[...]

PR #97, opened 2016-03-23
Issue 94 - support for custom namespaces

iuyiuy:
This should fix issue #94 - support for custom namespaces.


codedread:
Sorry for the delay here - would you mind syncing with our trunk so I can review this?

editor/sanitize.js
```diff
@@ -154,17 +153,16 @@ svgedit.sanitize.sanitizeSvg = function(node) {
       var attrName = attr.nodeName;
       var attrLocalName = attr.localName;
       var attrNsURI = attr.namespaceURI;
+
+      // Ignore attributes from custom namespaces
+      var splitNS = attrName.match(/([^:]+):/);
+      if(splitNS && splitNS[1].toUpperCase() in svgedit.ignoredNS) continue;
```
flintobrien: Regarding ```toUpperCase()```
I could be wrong, but I believe namespace prefixes in SVG are case sensitive. If I'm right, the svgedit.getReverseNS() function in the current release is also wrong.

That's according to web forums and the only specs I could find. See:
> [SVG 6. Styling](https://www.w3.org/TR/SVG/styling.html#CaseSensitivity) — 6.7 Case sensitivity of property names and values
> [CSS Namespaces Module](https://www.w3.org/TR/2011/REC-css3-namespace-20110929/) — 3.3. Declaring Prefixes

flintobrien: Perhaps this is this what's going on...
The svgedit.NS.* properties are upper case because they are constants. The rest of the svgedit codebase assumes namespaces prefixes are always lower case (incorrectly?), so getReverseNS() calls toLower() on each property name. Given all that, line 159 above converts the namespace prefix on each attribute to the upper-case object property name.

Given all that, this code works the same as the rest of the codebase.

editor/svgedit.js
```diff
@@ -10,18 +10,20 @@ svgedit = {
 	NS: {
 		HTML: 'http://www.w3.org/1999/xhtml',
 		MATH: 'http://www.w3.org/1998/Math/MathML',
-		SE: 'http://svg-edit.googlecode.com',
 		SVG: 'http://www.w3.org/2000/svg',
 		XLINK: 'http://www.w3.org/1999/xlink',
 		XML: 'http://www.w3.org/XML/1998/namespace',
 		XMLNS: 'http://www.w3.org/2000/xmlns/' // see http://www.w3.org/TR/REC-xml-names/#xmlReserved
+	},
+	ignoredNS: {
```
flintobrien: As a constant, ingnoredNS should probably be NS_IGNORE.

iuyiuy:
I've now synced the branch.

Reg. case sensitivity: I think I noticed that the rest of the code was treating it as case insensitive and didn't question it. But I think you're right, in that it should be case sensitive. I've fixed this in the branch now.

Reg. ignoredNS vs. NS_IGNORE: It's currently implemented so namespaces to be ignored can be defined in config.js and this list is merged with that list. So it's not entirely a constant.

editor/svgedit.js
```diff
@@ -8,21 +8,23 @@
...
 	// common namepaces constants in alpha order
 	NS: {
-		HTML: 'http://www.w3.org/1999/xhtml',
-		MATH: 'http://www.w3.org/1998/Math/MathML',
-		SE: 'http://svg-edit.googlecode.com',
-		SVG: 'http://www.w3.org/2000/svg',
-		XLINK: 'http://www.w3.org/1999/xlink',
-		XML: 'http://www.w3.org/XML/1998/namespace',
-		XMLNS: 'http://www.w3.org/2000/xmlns/' // see http://www.w3.org/TR/REC-xml-names/#xmlReserved
+		html: 'http://www.w3.org/1999/xhtml',
```
codedread: I don't understand why these needed to be changed.  Conventionally uppercase things are constants (which namespaces are)

distler: +1 for leaving namespaces as upper-case constants. (Adheres to convention *and* will make this pull-request a *lot* shorter.)

editor/sanitize.js
```diff
@@ -154,17 +153,17 @@ svgedit.sanitize.sanitizeSvg = function(node) {
       var attrName = attr.nodeName;
       var attrLocalName = attr.localName;
       var attrNsURI = attr.namespaceURI;
+
+      // Ignore attributes from custom namespaces and their definitions
+      var splitNS = attrName.match(/([^:]+):(.*$)/);
+      if(splitNS && splitNS[1] in svgedit.ignoredNS) continue;
+      if(splitNS && splitNS[1] == 'xmlns' && splitNS[2] in svgedit.ignoredNS) continue;
```
distler: Umh. No.

This is *exactly* the sort of thing I was complaining about.

```
xmlns:se="http://www.w3.org/1999/xhtml" se:onload=..."
```

(and variations on that theme) would now allow a user to enter arbitrary javascript.

You need, *at a minimum*, to check not only that the prefix is one of the keys in the `ignoredNS` hash, but that the *value* is correct.

Better yet (to continue beating a very dead horse) would be an explicit whitelist of elements and attributes. But, if we're going to whitelist a whole namespace, we at least need to do that correctly.

iuyiuy:
Well it was in response to @flintobrien, who pointed out that namespaces in xml and thus svg are case sensitive. It wasn't really part of issue #94, but I decided to fix it, while I was working on that part of it anyways. Adding a case insensitive conf setting, only to later change it, would also be bad.

But these values aren't constants. First of all constants don't actually exist in JS, but we could pretend and that does make sense, in order to "protect" their use, even if only enforced by code reviewers. But these are the keys in the object/hash/dictionary, which the 'constant' is made up of. These keys represent a lower case text, because most namespaces are referred to in lower case. If someone wanted to make a LaTeX namespace, they could and even use that particular casing, in order to avoid confusion with things with other casing. The actual name of the constant is NS and that part is upper case, after all.

Would anyone argue against this? The only argument I don't think I addressed, was that the pull request would be shorter without this change. But that's obviously not a good reason if it's an improvement and even kinda necessary, in order to not have a conf setting, which will be changed at some point.

---
Reg. the security issue of allowing to ignore a full custom namespace, defined in config.js. @distler: I can't see your example working in any way. I couldn't even get html:onload or svg:onload to work in Chrome in an svg, even if onload works. And even if they worked, they'd only work because they were namespaces that the browser recognized. The browser wouldn't know "se" as a namespace and thus wouldn't interpret se:onload and execute it. I'd love to see an example where this could be used in any way for something bad.

I could see an argument for not having "se" in the default ignored/whitelisted namespace list and specifying all allowed attributes. With that particular namespace, the reference implementation is here after all. But a site-defined (read: defined in config.js) namespace, should be possible to have without needing to keep your config.js in sync with the namespace. Particularly if this is a namespace the browser just ignores and leaves as regular ignored attributes (and tags, btw).

codedread:
Part 1:  The NS.SVG "constants" are more like Java enums.  That's the
idea.  They are symbols.
Part 2:  http://codedread.com/testbed/circle.svg

codedread:
FWIW, I don't care too strongly about the upper-case part and I didn't look
very closely at your patch.  I acknowledge that making them lowercase might
have made something easier that I don't quite understand.

iuyiuy:
@codedread: You're right. The url for the namespace is the part which identifies how it should be interpreted in the browser or if it should just be ignored. Because the url relates to the definition and html/svg/se are just names or aliases locally in the xml document. So as I said, the html and svg namespaces (defined by the URLs) can invoke scripting. Even if they're called "se" in the svg document. If you define xmlns:html="https://github.com/SVG-Edit/svgedit" in svg, the html:script tag won't work, but set it to xmlns:html="http://www.w3.org/1999/xhtml" and suddenly it works again.

So I did make a mistake. Thanks for guiding me there. I'm gonna change it to allow namespaces based on the url. This means we could allow them regardless of their local name/alias in the svg. Which also means, we could keep the name for them in the NS file as before and just have an array of site-defined namespace URLs to ignore, regardless of their namespace alias in the svg.

What do you say? Isn't that the right way? Should i revert the namespace names, so they're back to being in upper case? They'll be ignored in the actual code usage anyways and only used to refer to which is which.

iuyiuy:
I've now reverted the change to the variables, so the namespace prefix/alias defined for the built-in or whitelisted namespaces are all upper-case again. I've also moved the se namespace back to this list.

And now the ignored namespaces are detected by the URI of the namespaces (a.k.a. their definition) and not their prefix/alias. Which means you can ignore https://example.com/my-personal-extension regardless of what you call it. You could define it with xmlns:foo and then have foo:bar nodes/attributes. You can also define it with xmlns:html and have html:script, but if you've defined it to use a namespace that the browser doesn't recognize, it won't interpret it as script or anything else. So this isn't a security problem that way. And it's only allowed if it's set in the config.js file.

I think the built-in whitelist should be uri-based instead of alias-based as well. But I didn't include that here, to avoid scope creep in this pull request.

So are there any comments, issues, concerns or anything else about the pull request now? Did i miss something? I'm sorry it ended up being a messy bunch of commits. The end result is important though.

codedread:
Thanks KB - I've been pretty busy to look at this.  If Jacques or others
have time, that'd be great - otherwise I'm likely to get to this by the
weekend...

iuyiuy:
Great. Thank you.

iuyiuy:
I've made a small change to support nodes with a custom namespace too. It will now treat any nodes with ignored namespaces as a group you can't ungroup and any svg elements in it will be shown and aren't editable.

iuyiuy:
I noticed an issue in resizing elements with custom namespaces. It should work better now.

iuyiuy:
I've also fixed an issue in ext-connector.js for this feature. This included a side-effect of improving/simplifying getBBoxWithTransform() quite a bit.

evssraghavendra:
I tried to use the code from this tree and i see that there are different exceptions that are thrown.
svgutils.js:495 Uncaught TypeError: Cannot use 'in' operator to search for 'null' in undefined

if(elname === 'use' || elem.prefix in svgedit.ignoredNSUsedAlias) 

Is there some thing this is missed during checin process?
image

iuyiuy:
@evssraghavendra: This is now fixed in the latest commit. The svgedit.ignoredNSUsedAlias was only initialized in sanitizeSvg(), which is run when you load a drawing programmatically with setSvgString(), but not when opening a new drawing. It's now initialized when the drawing is initialized. This should cover the bases.

evssraghavendra:
What is the mechanism to discuss about this tool other than logging an issue.

iuyiuy:
This is the pull request for this implementation. So if you want to discuss this implementation, it would be here. For the generic problem that this is trying to solve, the issue is issue https://github.com/SVG-Edit/svgedit/issues/94.

iuyiuy:
I found out that the custom namespaces weren't applied to nodes properly, which is why they were visible in the browser. I've now removed the faulty handling of these nodes, since if it's not visible, it's not clickable. So now both nodes and attributes from custom namespaces are safely ignored by svg-edit.

Then i've added a new se attribute called se:external, which you can set to "1" on a g node. If this is set, the group won't be ungroup-able and transforms won't be applied to children of the group. This is so that you can have groups that are treated as a regular element that you can set your own toolbar/etc. type controls for. And then you can use regular svg elements as its children, to make it display as you want in svg-edit.

iuyiuy:
Is there anything holding this pull request back? Are there any issues or concerns with it?

brettz9:
I'd need to test, but from my perspective, the thing holding it back is it unfortunately it needing rebasing so we can test again... That should greatly expedite the chances and speed by which I may be able to review and pull...

editor/svgutils.js
```diff
@@ -816,68 +820,21 @@ function bBoxCanBeOptimizedOverNativeGetBBox(angle, hasMatrixTransform) {
...
+    var bgRect = $('#canvasBackground>rect')[0].getBoundingClientRect();
+    var elRect = elem.getBoundingClientRect();
+    var zoom = svgCanvas.getZoom();
+
+    return {
+        x: (elRect.left - bgRect.left) / zoom,
+        y: (elRect.top - bgRect.top) / zoom,
+        width: elRect.width / zoom,
+        height: elRect.height / zoom,
+    };
```
brettz9: This fix looks a lot shorter than the previous code... Can you summarize how this improves upon the previous code without losing anything?

jfhenon:
Hello, I've been recently added to the maintainer team. I see that this PR is quite old, had many commits but was never merged in the master branch. I think we can put some efforts into this PR but I would like to know if @iuyiuy is still interested and willing to help as he was the main (actually the only) contributor to this PR?

distler:
I don't recall whether the security issues raised were satisfactorily addressed. If you decide to resurrect this PR, that's something to check carefully.

jfhenon:

PR #103, opened 2016-04-24
GetBBox performance improvements

flintobrien:
We just updated our project to the latest svg-edit (last update was three years ago). We've seen some major performance problems when performing selections in Chrome, Firefox, and Safari. With a large SVG document (270 top-level elements), select-all takes over 1 minute. This is especially evident when the top-level selections have matrix transforms.  When there are a few elements, we don't see the problem.

See [issue 26](https://github.com/SVG-Edit/svgedit/issues/26). We see the problems in the demo version pointed to from the svgedit page.

Chrome's profiler points to the native getBBox() taking all the time.

**The problems are due to:**

1) Both the thumbnail preview window and the prevalence of elements with matrix transforms (in our sample document). When getStrokedBBox/getCheckedBBox() sees a matrix transform, it calculates the BBox by enclosing a clone of the element and adding it to the DOM, calling getBBox(), then removing it. Without the transforms, there's no DOM change, so it's faster. If you delete the thumbnail and keep the transforms, it's faster, but still slow.

2) The various getBBox utility functions call the native getBBox() function multiple times for each element. It's at least 4 times for each top-level element when performing a rubber band selection.

3) svgcanvas: addToSelection() calls svgedit.utilities.getBBox(), throws the bbox away, then calls selectorManager.requestSelector() which calls svgedit.utilities.getBBox() again.

I've worked this week to refactor the related functions to calculate the BBox without updating the DOM and only call the native getBBox() once. I hope it's satisfactory. I've added a lot of tests and tested in Chrome, Firefox, and Safari on the latest OS X.

**Test Notes:**
* Please see the new performance test I added, svgutils_performance_test.html. I didn't add it to all_tests because I'm not sure what performance is expected for all platforms and devices.
* There are two new tests in svgedit.utilities_bbox that fail in Safari only. The last case in each deals with a rotated ellipse. I believe the BBox algorithm is incorrectly including the bezier control points to calculate the bounding box. i believe the Safari result is the correct one, but I'm not sure what is expected.
  * Test getBBoxWithTransform and a rotation transform.
  * Test getBBoxWithTransform with rotation and matrix transforms 

We're using svgedit for a schematic design tool (electric utility one-line diagrams).


flintobrien:
Here are the performance improvement times with various browsers. The old code takes 4000 milliseconds to select-all and the new code is 34 milliseconds.

Test case test/svgutils_performance_test.html run on Macbook 2011 8MB RAM OS X El Capitan 10.11.4.
"Pass2" (below) is what you normally see when using svgedit to start a rubber band selection. Note: starting a rubber band selection calls getBBox on all objects.

#### Chrome
* Before Optimization
  * Pass1 svgCanvas.getStrokedBBox total ms 4,218, ave ms 41.0,   min/max 37 51
  * Pass2 svgCanvas.getStrokedBBox total ms 4,458, ave ms 43.3,   min/max 32 63
* Optimized Code
  * Pass1 svgCanvas.getStrokedBBox total ms 1,112, ave ms 10.8,   min/max 9 20
  * Pass2 svgCanvas.getStrokedBBox total ms    34, ave ms  0.3,   min/max 0 20

#### Firefox
* Before Optimization
  * Pass1 svgCanvas.getStrokedBBox total ms 3,794, ave ms 36.8,   min/max 33 48
  * Pass2 svgCanvas.getStrokedBBox total ms 4,049, ave ms 39.3,   min/max 28 53
* Optimized Code
  * Pass1 svgCanvas.getStrokedBBox total ms   104, ave ms 1.0,   min/max 0 23
  * Pass2 svgCanvas.getStrokedBBox total ms    71, ave ms 0.7,   min/max 0 23

#### Safari
* Before Optimization
  * Pass1 svgCanvas.getStrokedBBox total ms 4,840, ave ms 47.0,   min/max 45 62
  * Pass2 svgCanvas.getStrokedBBox total ms 4,849, ave ms 47.1,   min/max 34 62
* Optimized Code
  * Pass1 svgCanvas.getStrokedBBox total ms    42, ave ms 0.4,   min/max 0 23
  * Pass2 svgCanvas.getStrokedBBox total ms    17, ave ms 0.2,   min/max 0 23

If you want to run the performance test to see 'Before Optimization' yourself, make the following two edits to the new code.
1 svgedit.utilities.getStrokedBBox - change "if( elems.length === 1)" to "if( false && elems.length === 1)"
2 svgedit.utilities.getBBoxWithTransform - uncomment 'Old technique that was very slow'

codedread:
Wow - this sounds awesome - I'm hoping to take a look at this today or tonight.

flintobrien:
I'm happy to do a phone call and walk through the changes. Whatever I can
do to help.
Regards,
Flint
919.260.8411

editor/select.js
```diff
@@ -162,7 +165,10 @@ svgedit.select.Selector.prototype.resize = function() {
 	m.e *= current_zoom;
 	m.f *= current_zoom;
 
-	var bbox = svgedit.utilities.getBBox(selected);
+	if (!bbox)
+		bbox = svgedit.utilities.getBBox(selected);
```
codedread: Can you please put curly braces around this if?

editor/svgcanvas.js
```diff
@@ -601,140 +601,9 @@ var getIntersectionList = this.getIntersectionList = function(rect) {
...
-			var offset = getOffset(elem);
-			max_x = Math.max(max_x, cur_bb.x + cur_bb.width + offset);
-			max_y = Math.max(max_y, cur_bb.y + cur_bb.height + offset);
-		}
-	});
-	
-	full_bb.width = max_x - min_x;
-	full_bb.height = max_y - min_y;
-	return full_bb;
+	return svgedit.utilities.getStrokedBBox( elems, addSvgElementFromJson, pathActions)
```
codedread: nit: please remove the space before elems

editor/svgcanvas.js
```diff
@@ -6576,17 +6447,15 @@ this.setSegType = function(new_type) {
...
 		return;
 	}
-	
-	if (!getBBox) {
-		var batchCmd = new svgedit.history.BatchCommand('Convert element to Path');
-	}
-	
-	var attrs = getBBox?{}:{
+	if( getBBox) {
+		return svgedit.utilities.getBBoxOfElementAsPath( elem, addSvgElementFromJson, pathActions)
```
codedread: nit: remove space in from of elem

editor/svgcanvas.js
```diff
@@ -6576,17 +6447,15 @@ this.setSegType = function(new_type) {
...
 		});
 		return;
 	}
-	
-	if (!getBBox) {
-		var batchCmd = new svgedit.history.BatchCommand('Convert element to Path');
-	}
-	
-	var attrs = getBBox?{}:{
+	if( getBBox) {
```
codedread: nit: pls remove space in front of getBBox

editor/svgutils.js
```diff
@@ -529,19 +529,387 @@ svgedit.utilities.getBBox = function(elem) {
 	return ret;
 };
 
-// Function: svgedit.utilities.getRotationAngle
-// Get the rotation angle of the given/selected DOM element
+svgedit.utilities.getPathDFromSegments = function( pathSegments) {
```
codedread: This is the 4th time I've seen a leading space in front of the first argument, so maybe it's a coding convention I'm not familiar with?

editor/svgutils.js
```diff
@@ -529,19 +529,387 @@ svgedit.utilities.getBBox = function(elem) {
...
+	$.each(pathSegments, function(j, seg) {
+		var i;
+		var l = seg[0], pts = seg[1];
+		d += l;
+		for (i = 0; i < pts.length; i+=2) {
+			d += (pts[i] +','+pts[i+1]) + ' ';
+		}
+	});
+
+	return d
```
codedread: missing semicolon

editor/svgutils.js
```diff
@@ -529,19 +529,387 @@ svgedit.utilities.getBBox = function(elem) {
...
-// Get the rotation angle of the given/selected DOM element
+svgedit.utilities.getPathDFromSegments = function( pathSegments) {
+	var d = '';
+
+	$.each(pathSegments, function(j, seg) {
+		var i;
+		var l = seg[0], pts = seg[1];
+		d += l;
+		for (i = 0; i < pts.length; i+=2) {
+			d += (pts[i] +','+pts[i+1]) + ' ';
```
codedread: It's been awhile since I looked at SVG code, but this assumes that the segments are only two coordinates a piece, right?  Aren't there some segments that are multiple numbers (arcs and the like)

editor/svgutils.js
codedread: Can you add a comment to this function?

editor/svgutils.js
```diff
@@ -529,19 +529,387 @@ svgedit.utilities.getBBox = function(elem) {
...
+		}
+	});
+
+	return d
+}
+
+svgedit.utilities.getPathDFromElement = function( elem) {
```
codedread: Add a comment

editor/svgutils.js
```diff
@@ -529,19 +529,387 @@ svgedit.utilities.getBBox = function(elem) {
...
-// elem - Optional DOM element to get the angle for
+// elem - The DOM element to be probed
+// addSvgElementFromJson - Function to add the path element to the current layer. See canvas.addSvgElementFromJso
+// pathActions - If a transform exists, pathActions.resetOrientation() is used. See: canvas.pathActions.
+//
+// Returns:
+// The resulting path's bounding box object.
+svgedit.utilities.getBBoxOfElementAsPath = function(elem, addSvgElementFromJson, pathActions) {
+
+	var attrs = {}
```
codedread: missing semicolon

editor/svgutils.js
```diff
@@ -529,19 +529,387 @@ svgedit.utilities.getBBox = function(elem) {
...
+
+	// any attribute on the element not covered by the above
```
codedread: Comment doesn't seem to apply

editor/svgutils.js
```diff
@@ -529,19 +529,387 @@ svgedit.utilities.getBBox = function(elem) {
...
+// history - see svgedit.history
+// addCommandToHistory - see canvas.addCommandToHistory
+//
+// Returns:
+// The converted path element or null if the DOM element was not recognized.
+svgedit.utilities.convertToPath = function(elem, attrs, addSvgElementFromJson, pathActions, clearSelection, addToSelection, history, addCommandToHistory) {
+
+	var batchCmd = new history.BatchCommand('Convert element to Path');
+
+	// any attribute on the element not covered by the above
```
codedread: Comment doesn't seem to apply now

editor/svgutils.js
```diff
@@ -529,19 +529,387 @@ svgedit.utilities.getBBox = function(elem) {
...
+	return bb;
+}
+
+// Function: convertToPath
+// Convert selected element to a path.
+//
+// Parameters:
+// elem - The DOM element to be converted
+// attrs - Apply attributes to new path. see canvas.convertToPath
+// addSvgElementFromJson - Function to add the path element to the current layer. See canvas.addSvgElementFromJso
```
codedread: missing a trailing 'n' :-)

editor/svgutils.js
```diff
@@ -529,19 +529,387 @@ svgedit.utilities.getBBox = function(elem) {
...
+	});
+};
+
+// Function: getBBoxOfElementAsPath
+// Get the BBox of an element-as-path
 //
 // Parameters:
-// elem - Optional DOM element to get the angle for
+// elem - The DOM element to be probed
+// addSvgElementFromJson - Function to add the path element to the current layer. See canvas.addSvgElementFromJso
```
codedread: It's a little weird to me to be passing in this function... the utility module shouldn't know about the canvas, it's just a module that provides utilities ideally.

editor/svgutils.js
```diff
@@ -529,19 +529,387 @@ svgedit.utilities.getBBox = function(elem) {
...
+// pathActions - If a transform exists, pathActions.resetOrientation() is used. See: canvas.pathActions.
```
codedread: Ugh - this reminds me of a big TODO I had a long time ago where I wanted to pull all of canvas.pathActions into a separate module.  That 1000 lines of code shouldn't be sitting in svgcanvas.js :-/

(I'm not asking you to do any of that, it's just an FYI)

editor/svgutils.js
```diff
@@ -529,19 +529,387 @@ svgedit.utilities.getBBox = function(elem) {
...
+//
+// Parameters:
+// elem - The DOM element to be converted
+// attrs - Apply attributes to new path. see canvas.convertToPath
+// addSvgElementFromJson - Function to add the path element to the current layer. See canvas.addSvgElementFromJso
+// pathActions - If a transform exists, pathActions.resetOrientation() is used. See: canvas.pathActions.
+// clearSelection - see canvas.clearSelection
+// addToSelection - see canvas.addToSelection
+// history - see svgedit.history
+// addCommandToHistory - see canvas.addCommandToHistory
```
codedread: Same comment about all these svgcanvas-specific functions you are passing into a "utility" function.  It means this isn't really a utility function after all.

I don't want to derail the huge performance wins you've made with this patch, it's great - but sending functions into this function to be called just ultimately makes the code harder to understand later.  Maybe think about it in future patches?

codedread:
Wow - thanks so much for this patch - I didn't look through the tests, just sort of skimmed through them.  Mostly I have concerns about passing svgcanvas functions into module functions - it's a bit of a hack, but frankly it's my fault for leaving the code in a half-migrated state where some of the code lives in separate modules and most of the code is still sitting in svgcanvas.js in a huge ball of spaghetti :)

editor/svgutils.js
flintobrien: I'll be fixing the spaces and semicolons; but I wanted to give feedback on this for now...

**Not adding to DOM:**
I've had issues in the past where, if the element is not attached to the DOM, the native getBBox returns zero width and height (in Chrome specifically). I'm also worried that it might work on some browser versions/platforms, but not others.

**Why create a path at all**
getBBoxWithTransform calls getBBoxOfElementAsPath when there is a simple shape with a transform. I assume it's doing this because (for example) a _rounded_ rect rotated 45 is not as wide as a _non-rounded_ rect rotated 45.

When it's not a simple shape, the new code gets the native bbox, then does matrix transforms on the bbox. This is because all the current browsers return the bbox without top-level transforms (and why the old code wrapped the transform inside a new group, then got the bbox).

Oddly, I tested the current release and the new code. The selections are the same before and after in Chrome and Safari. Firefox does seem to calculate a tighter bbox in the current released code (though with performance issues).

Here's an example with rotated ellipse and rotate g. Try rubber band selection starting from below to see if the bottom corner of the circle selection is part of the rubber band hit. The only advantage for the old code is with Firefox.
```
  <g id="svg_3" transform="rotate(45, 100, 100)">
   <ellipse fill="#56ff56" stroke-width="5" stroke-dasharray="null" stroke-linejoin="null" stroke-linecap="null" cx="100" cy="100" rx="50" ry="50" id="svg_1"/>
   <ellipse fill="#00bf5f" stroke-width="5" stroke-dasharray="null" stroke-linejoin="null" stroke-linecap="null" cx="116" cy="100" rx="20" ry="20" id="svg_2"/>
  </g>
  <ellipse transform="rotate(45, 300, 100)" ry="50" rx="50" id="svg_4" cy="100" cx="300" stroke-width="5" stroke="#000000" fill="#ff0000"/>
```

**Injected Functions**
In general, I like injected _services_ because it makes it easier to create unit tests.

I agree it would be much better if utility functions didn't depend on canvas. This is a refactoring of the original code in canvas.convertToPath. I broke it out into two functions, changing convertToPath internally, but not updating references _using_ convertToPath. I wanted to get feedback before taking it further.

**Next Steps**
Interestingly, the new code doesn't affect the selection handles bbox you see when you click on an object.

I think the slightly optimized bbox (in Firefox only) is not worth the performance hit.

We could remove the getBBoxOfElementAsPath calls in getBBoxWithTransform if we can live with the sub-optimal bbox hit-test for rubber band selection. Given this, we can remove both injected functions from the utilities.

**Future**
We could still optimize some shapes manually in the future. For example, it would be relatively easy to calculate the actual width of a rotated rounded rect. A more advanced transform would be trickier.

There is a getBoundingClientRect() that calculates the bbox with the transform which sounds like exactly what we need; however, it would have to be translated back to SVG space and I didn't have time to investigate that further.

flintobrien:
I copied this discussion here because it seems to be hidden with my latest commit (says 'comment on outdated diff'). 

I believe I've fixed the formatting with the latest commit.

**Not adding to DOM:**
I've had issues in the past where, if the element is not attached to the DOM, the native getBBox returns zero width and height (in Chrome specifically). I'm also worried that it might work on some browser versions/platforms, but not others.

**Why create a path at all**
getBBoxWithTransform calls getBBoxOfElementAsPath when there is a simple shape with a transform. I assume it's doing this because (for example) a _rounded_ rect rotated 45 is not as wide as a _non-rounded_ rect rotated 45.

When it's not a simple shape, the new code gets the native bbox, then does matrix transforms on the bbox. This is because all the current browsers return the bbox without top-level transforms (and why the old code wrapped the transform inside a new group, then got the bbox).

Oddly, I tested the current release and the new code. The selections are the same before and after in Chrome and Safari. Firefox does seem to calculate a tighter bbox in the current released code (though with performance issues).

Here's an example with rotated ellipse and rotate g. Try rubber band selection starting from below to see if the bottom corner of the circle selection is part of the rubber band hit. The only advantage for the old code is with Firefox.
```
  <g id="svg_3" transform="rotate(45, 100, 100)">
   <ellipse fill="#56ff56" stroke-width="5" stroke-dasharray="null" stroke-linejoin="null" stroke-linecap="null" cx="100" cy="100" rx="50" ry="50" id="svg_1"/>
   <ellipse fill="#00bf5f" stroke-width="5" stroke-dasharray="null" stroke-linejoin="null" stroke-linecap="null" cx="116" cy="100" rx="20" ry="20" id="svg_2"/>
  </g>
  <ellipse transform="rotate(45, 300, 100)" ry="50" rx="50" id="svg_4" cy="100" cx="300" stroke-width="5" stroke="#000000" fill="#ff0000"/>
```

**Injected Functions**
In general, I like injected _services_ because it makes it easier to create unit tests.

I agree it would be much better if utility functions didn't depend on canvas. This is a refactoring of the original code in canvas.convertToPath. I broke it out into two functions, changing convertToPath internally, but not updating references _using_ convertToPath. I wanted to get feedback before taking it further.

**Next Steps**
Interestingly, the new code doesn't affect the selection handles bbox you see when you click on an object.

I think the slightly optimized bbox (in Firefox only) is not worth the performance hit.

We could remove the getBBoxOfElementAsPath calls in getBBoxWithTransform if we can live with the sub-optimal bbox hit-test for rubber band selection. Given this, we can remove both injected functions from the utilities.

**Future**
We could still optimize some shapes manually in the future. For example, it would be relatively easy to calculate the actual width of a rotated rounded rect. A more advanced transform would be trickier.

There is a getBoundingClientRect() that calculates the bbox with the transform which sounds like exactly what we need; however, it would have to be translated back to SVG space and I didn't have time to investigate that further.

flintobrien:
On reflection....
If you're trying to line things up, stay on grid, or do precise resizing of a transformed element, it is paramount that the bbox be accurate (both the bounding rect and selection handles).

codedread:
Ok, I see a lot of the other comments have been cleaned up - if you think this is ready, I will merge it... just let me know.

flintobrien:
Thanks for the reviews. I think it's good to go with my last commit, [Optimized getBBoxWithTransform when rotation multiple of 90](https://github.com/SVG-Edit/svgedit/pull/103/commits/3230520d679a334cc1938dfbb5972783f63fd9b9).

Issue #104, opened 2016-04-25
calling canvas.createLayer() programatically doesn't update layers panel


flintobrien:
I'm happy to fix this, just looking to narrow down possible options....

### Problem
For our application; on startup of svgedit, we rename "Layer 1" and create a second layer. The layers panel is not repopulated and still shows one layer, "Layer 1" even though the document has been updated correctly.

### Analysis
createLayer does issue `call('changed', [new_layer])`, but the layers panel is looking for change to `svg#svgcontent`, not the `<g>` element of the new layer.

Here's the relevant code in svg-editor.js
```
var elementChanged = function(win, elems) {
  ...
  if (elem && elem.tagName === 'svg') {
    populateLayers();
    updateCanvas();
  }
  ...
}
```

### Possible fixes
1) Add a `class="layer"` to the layer group element to identify layers. Then elementChanged can easily identify a layer change. This is clean but not backwards compatible for old documents (unless we add code to update them on load).
2) In elementChanged, add an additional test for elem.parentNode being `svg` with `id === 'svgcontent'` to identify layer changes.
3) `svgedit.draw.Drawing.prototype.identifyLayers` could issue a more direct broadcast message to notify the layers panel that the layers have changed. I don't know the repercussions of this.
4) Other ideas?

From what I currently know, I'd recommend option 2. If I was designing from scratch, I'd pick 1.

codedread:
Ok, I will have to try look at the code and think more about this (maybe tonight or tomorrow).  I have a gut preference for suggestion #1 with the caveat that documents re-loaded back in would get the class="layer" added like you suggest

PR #106, opened 2016-04-29
Update layers panel on programatic call to createLayer.

flintobrien:
This is intended to fix [issue 104](https://github.com/SVG-Edit/svgedit/issues/104). It adds class=“layer” to each layer element, so calling createLayer programmatically is easily detected in elementChanged and the layers panel is always updated appropriately. It's backwards compatible with legacy drawings because it auto-adds the class to legacy drawings.

**Note:** [coderead](https://github.com/codedread) seemed to prefer this solution, but wasn't committed. I decided to try it out and see how it went. I think it worked out well; so I decided to do a pull request. See commit notes below.

### Problem
For our application; on startup of svgedit, we rename "Layer 1" and create a second layer. The layers panel is not repopulated and still shows one layer, "Layer 1" even though the document has been updated correctly.

### Analysis
createLayer does issue `call('changed', [new_layer])`, but the layers panel is looking for change to `svg#svgcontent`, not the `<g>` element of the new layer.

Here's the relevant code in svg-editor.js
```
var elementChanged = function(win, elems) {
  ...
  if (elem && elem.tagName === 'svg') {
    populateLayers();
    updateCanvas();
  }
  ...
}
```

### Commit Notes
Add class=“layer” to each layer element. This happens when calling createLayer() or any call to identifyLayers(). This happens with new drawings and when opening legacy drawings, so it’s fully backwards compatible.

svg-editor.js elementChanged() looks for g.layer as an addition test when calling populateLayers(). This updates the layers panel.

Added tests for class=“layer” added to draw_test.html.

Fixed Firefox exception in draw_test.html. It was failing in native getBBox() because, in the test, the SVG element was not attached to the DOM.

### DIFF
--- a/editor/draw.js
+++ b/editor/draw.js
@@ -20,6 +20,8 @@ if (!svgedit.draw) {
 }
 // alias
 var NS = svgedit.NS;
+var LAYER_CLASS = svgedit.LAYER_CLASS;
+var LAYER_CLASS_REGEX = svgedit.LAYER_CLASS_REGEX;
 
 var visElems = 'a,circle,ellipse,foreignObject,g,image,line,path,polygon,polyline,rect,svg,text,tspan,use'.split(',');
 
@@ -29,6 +31,21 @@ var RandomizeModes = {
 	NEVER_RANDOMIZE: 2
 };
 var randomize_ids = RandomizeModes.LET_DOCUMENT_DECIDE;
+
+/**
+ * Add class 'layer' to the element
+ *
+ * Parameters:
+ * @param {SVGGElement} elem - The SVG element to update
+ */
+function addLayerClass(elem) {
+	var classes = elem.getAttribute('class')
+	if (classes === null || classes === undefined || classes.length === 0) {
+		elem.setAttribute('class', LAYER_CLASS);
+	} else if (! LAYER_CLASS_REGEX.test(classes)) {
+		elem.setAttribute('class', classes + ' ' + LAYER_CLASS);
+	}
+}
 
 /**
  * This class encapsulates the concept of a layer in the drawing
@@ -386,6 +403,7 @@ svgedit.draw.Drawing.prototype.identifyLayers = function() {
 					a_layer = child;
 					svgedit.utilities.walkTree(child, function(e){e.setAttribute("style", "pointer-events:inherit");});
 					a_layer.setAttribute("style", "pointer-events:none");
+					addLayerClass(a_layer)
 				}
 				// if group did not have a name, it is an orphan
 				else {
@@ -407,21 +425,35 @@ svgedit.draw.Drawing.prototype.identifyLayers = function() {
 		// TODO(codedread): What about internationalization of "Layer"?
 		while (layernames.indexOf(("Layer " + i)) >= 0) { i++; }
 		var newname = "Layer " + i;
-		a_layer = svgdoc.createElementNS(NS.SVG, "g");
-		var layer_title = svgdoc.createElementNS(NS.SVG, "title");
-		layer_title.textContent = newname;
-		a_layer.appendChild(layer_title);
+		a_layer = this._doCreateLayer(newname);
 		var j;
 		for (j = 0; j < orphans.length; ++j) {
 			a_layer.appendChild(orphans[j]);
 		}
-		this.svgElem_.appendChild(a_layer);
-		this.all_layers.push( [newname, a_layer] );
+		this.all_layers.push([newname, a_layer] );
 	}
 	svgedit.utilities.walkTree(a_layer, function(e){e.setAttribute("style", "pointer-events:inherit");});
 	this.current_layer = a_layer;
 	this.current_layer.setAttribute("style", "pointer-events:all");
 };
+
+	/**
+	 * Private function that actually creates a new top-level layer in the drawing
+	 * with the given name and sets the current layer to it.
+	 * @param {string} name - The given name
+	 * @returns {SVGGElement} The SVGGElement of the new layer, which is
+	 * also the current layer of this drawing.
+	 */
+	svgedit.draw.Drawing.prototype._doCreateLayer = function(name) {
+		var svgdoc = this.svgElem_.ownerDocument;
+		var new_layer = svgdoc.createElementNS(NS.SVG, "g");
+		addLayerClass(new_layer);
+		var layer_title = svgdoc.createElementNS(NS.SVG, "title");
+		layer_title.textContent = name;
+		new_layer.appendChild(layer_title);
+		this.svgElem_.appendChild(new_layer);
+		return new_layer;
+	};
 
 /**
  * Creates a new top-level layer in the drawing with the given name and 
@@ -431,12 +463,7 @@ svgedit.draw.Drawing.prototype.identifyLayers = function() {
  * also the current layer of this drawing.
 */
 svgedit.draw.Drawing.prototype.createLayer = function(name) {
-	var svgdoc = this.svgElem_.ownerDocument;
-	var new_layer = svgdoc.createElementNS(NS.SVG, "g");
-	var layer_title = svgdoc.createElementNS(NS.SVG, "title");
-	layer_title.textContent = name;
-	new_layer.appendChild(layer_title);
-	this.svgElem_.appendChild(new_layer);
+	var new_layer = this._doCreateLayer(name);
 	this.identifyLayers();
 	return new_layer;
 };
--- a/editor/draw.js
+++ b/editor/draw.js
@@ -39,7 +39,7 @@ var randomize_ids = RandomizeModes.LET_DOCUMENT_DECIDE;
  * @param {SVGGElement} elem - The SVG element to update
  */
 function addLayerClass(elem) {
-	var classes = elem.getAttribute('class')
+	var classes = elem.getAttribute('class');
 	if (classes === null || classes === undefined || classes.length === 0) {
 		elem.setAttribute('class', LAYER_CLASS);
 	} else if (! LAYER_CLASS_REGEX.test(classes)) {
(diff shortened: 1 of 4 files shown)


editor/svgedit.js
```diff
@@ -15,8 +15,10 @@ svgedit = {
 		XLINK: 'http://www.w3.org/1999/xlink',
 		XML: 'http://www.w3.org/XML/1998/namespace',
 		XMLNS: 'http://www.w3.org/2000/xmlns/' // see http://www.w3.org/TR/REC-xml-names/#xmlReserved
-	}
+	},
+	LAYER_CLASS: 'layer'
 };
+svgedit.LAYER_CLASS_REGEX = new RegExp('(\\s|^)' + svgedit.LAYER_CLASS + '(\\s|$)');
```
flintobrien: Not sure where to put these constants. 

Could also be one object directly on _svgedit_ instead of string and regex. Like the following.
```
svgedit = {
  ...
  LAYER: {  CLASS_NAME: 'layer }
};
svgedit.LAYER.CLASS_REGEX = new RegExp('(\\s|^)' + svgedit.LAYER.CLASS_NAME + '(\\s|$)');
```

editor/draw.js
```diff
@@ -31,6 +33,21 @@ var RandomizeModes = {
 var randomize_ids = RandomizeModes.LET_DOCUMENT_DECIDE;
 
 /**
+ * Add class 'layer' to the element
+ *
+ * Parameters:
+ * @param {SVGGElement} elem - The SVG element to update
+ */
+function addLayerClass(elem) {
+	var classes = elem.getAttribute('class')
```
codedread: missing semicolon

editor/draw.js
```diff
@@ -31,6 +33,21 @@ var RandomizeModes = {
...
+	if (classes === null || classes === undefined || classes.length === 0) {
+		elem.setAttribute('class', LAYER_CLASS);
+	} else if (! LAYER_CLASS_REGEX.test(classes)) {
+		elem.setAttribute('class', classes + ' ' + LAYER_CLASS);
+	}
```
codedread: What do you think about using classList to do this:

    elem.classList.add(LAYER_CLASS)

http://caniuse.com/#search=classlist says support started in IE10+

flintobrien: I was going to, but I saw [http://caniuse.com/#feat=classlist](http://caniuse.com/#feat=classlist) which said that IE doesn't support it. See footnote 1.

editor/draw.js
```diff
@@ -407,22 +425,36 @@ svgedit.draw.Drawing.prototype.identifyLayers = function() {
...
+	svgedit.draw.Drawing.prototype._doCreateLayer = function(name) {
+		var svgdoc = this.svgElem_.ownerDocument;
+		var new_layer = svgdoc.createElementNS(NS.SVG, "g");
+		addLayerClass(new_layer);
+		var layer_title = svgdoc.createElementNS(NS.SVG, "title");
+		layer_title.textContent = name;
+		new_layer.appendChild(layer_title);
+		this.svgElem_.appendChild(new_layer);
+		return new_layer;
+	};
```
codedread: This is starting to feel like the concept of a Layer deserves to be its own class.  I did a little sniffing around in this module and sure enough I made a Layer class and then never used it:  svgedit.draw.Layer exists at the top of this file :-/

Can you add the code to create a layer into the Layer constructor and then use it in some way?  Or is this too much refactoring to bite off right now?

I even see a TODO I added for the all_layers array:  "Turn this into an Array.<Layer>"

sigh... sorry about the mess!

editor/svg-editor.js
```diff
@@ -1835,6 +1835,15 @@ TODOS
 				});
 			};
 
+			/**
+			 * Test whether an element is a layer or not.
+			 * @param {SVGGElement} elem - The SVGGElement to test.
+			 * @returns {boolean} True if the element is a layer
+			 */
+			function isLayer(elem) {
+				return elem && elem.tagName === 'g' && svgedit.LAYER_CLASS_REGEX.test(elem.getAttribute('class'))
```
codedread: if you think classList is ok to use, then the latter part just becomes  elem.classList.contains(LAYER_CLASS)

editor/svgedit.js
codedread: If you use classList (DOMTokenList), then do you need the regex anymore?  CLASS_NAME can just be a static property on the svgedit.draw.Layer class.

editor/draw.js
flintobrien: I like it. I'll see what I can come up with.

editor/draw.js
codedread: My god!  I thought the days of worrying about IE support were long gone!  At least Edge supports it - sadly IE11 has a much higher marketshare than Edge still... *sniff*

flintobrien:
I've updated the branch with a fleshed out Layer class. Let me know what you think. I updated the draw_test.html and it's working, but I haven't done a lot of testing beyond that.

#### My thoughts:
I would prefer to remove all the draw.get/setCurrentLayerXXX in place of draw.getCurrentLayer().set/getXXX. Best case for now would leave them deprecated because, event we, are using those methods in our project. There is more code the way it is now.

The constants are still in svgedit.js. The main problem is that svg-editor.js needs the regex. If I moved it to the Layer class in draw.js, then svg-editor.js would depend on draw.js — creating a new large dependency. Pulling in large dependencies makes it harder to write unit tests.

Personally, I prefer smaller files than the current project. I would break out the Layer class into it's own file. This would also help with a smaller dependency for svg-editor.js. Thoughts?

editor/draw.js
```diff
@@ -111,17 +232,22 @@ svgedit.draw.Drawing = function(svgElem, opt_idPrefix) {
 	this.releasedNums = [];
 
 	/**
-	 * The z-ordered array of tuples containing layer names and <g> elements.
+	 * The z-ordered array of Layer objects. Each layer has a name
+	 * and group element.
 	 * The first layer is the one at the bottom of the rendering.
-	 * TODO: Turn this into an Array.<Layer>
-	 * @type {Array.<Array.<String, SVGGElement>>}
+	 * @type {Layer[]}
```
codedread: Closure Compiler comments are Array<Layer>

editor/draw.js
```diff
@@ -111,17 +232,22 @@ svgedit.draw.Drawing = function(svgElem, opt_idPrefix) {
...
 	 */
 	this.all_layers = [];
 
 	/**
+	 * Map of all_layers by name.
+	 * @type {Object.<string,Layer>}
```
codedread: Closure Compiler updated their parser awhile back and you can remove the . now.

editor/draw.js
```diff
@@ -111,17 +232,22 @@ svgedit.draw.Drawing = function(svgElem, opt_idPrefix) {
...
 	/**
-	 * The z-ordered array of tuples containing layer names and <g> elements.
+	 * The z-ordered array of Layer objects. Each layer has a name
+	 * and group element.
 	 * The first layer is the one at the bottom of the rendering.
-	 * TODO: Turn this into an Array.<Layer>
-	 * @type {Array.<Array.<String, SVGGElement>>}
+	 * @type {Layer[]}
 	 */
 	this.all_layers = [];
```
codedread: With the map, you could just remove this and just change loops to do `for (var layerName in this.layer_map.

flintobrien:
Just saw references to all_layers in svgcanvas.js (rename layer, etc.). Working on it.

editor/draw.js
```diff
@@ -344,101 +455,107 @@ svgedit.draw.Drawing.prototype.setCurrentLayer = function(name) {
...
-					this.all_layers.push( [name, child] );
-					a_layer = child;
-					svgedit.utilities.walkTree(child, function(e){e.setAttribute("style", "pointer-events:inherit");});
-					a_layer.setAttribute("style", "pointer-events:none");
-				}
-				// if group did not have a name, it is an orphan
-				else {
+					layer = new Layer(name, child);
+					this.all_layers.push(layer);
+					this.layer_map[ name] = layer;
```
codedread: nit: remove space before name

editor/draw.js
```diff
@@ -344,101 +455,107 @@ svgedit.draw.Drawing.prototype.setCurrentLayer = function(name) {
...
-		var j;
-		for (j = 0; j < orphans.length; ++j) {
-			a_layer.appendChild(orphans[j]);
-		}
-		this.svgElem_.appendChild(a_layer);
-		this.all_layers.push( [newname, a_layer] );
+		layer = new Layer(getNewLayerName(layernames), null, this.svgElem_);
+		layer.appendChildren(orphans);
+		this.all_layers.push(layer);
+		this.layer_map[ name] = layer;
```
codedread: same

editor/draw.js
```diff
@@ -31,31 +33,150 @@ var RandomizeModes = {
...
+		return undefined;
+	}
+	var svgdoc = svgElem.ownerDocument;
+	var new_layer = svgdoc.createElementNS(NS.SVG, "g");
+	var layer_title = svgdoc.createElementNS(NS.SVG, "title");
+	layer_title.textContent = name;
+	new_layer.appendChild(layer_title);
+	svgElem.appendChild(new_layer);
+	return new_layer;
+}
```
codedread: This is a little confusing.  We have a createLayer() function, then we have a Layer constructor which calls it, then there is a Drawing.createLayer().  It seems like a Layer should know how to create itself, can we make it part of the constructor?

If that's too much to refactor again, just say so - I like this patch overall and am ready to merge it.

codedread:
Looks pretty good!  Thanks for sticking with this and improving things :)

I would love to have layer in a separate file and totally agree that smaller files are better for testability, comprehension, etc.  That's why I started trying to pull apart svgcanvas.js years ago - I just never completed it.

editor/draw.js
flintobrien: Layers are ordered but referenced externally by name, so we need both container types depending on which function is called. We can have one container and always build the other on each function call, but it's less code, and I think cleaner code, to keep both around.

I'll add a comment in the code about this.

codedread: Good point - ok, I'm going to merge this pull request now.

codedread:
Thanks again - you rock!

editor/draw.js
flintobrien: Wouldn't work without the .

flintobrien:
I don't think you saw my note. I found some references in to drawing.all_layers and drawing.current_layer in svgcanvas.js. I just pushed up a fix, but it has more stuff that you request and some new ideas from me that I'ld like to get your take on. I'll write up a separate comment about it.

My preference would be to revert the merge to master for now, until I get your feedback.

flintobrien:
Latest [commit](https://github.com/gec/svgedit/commit/899853c963f1d64b51e547827ed3b69c1c317165) with fix for Canvas referencing Drawing internals...

Moved Layer class to layer.js. Question: The class is referenced off of svgedit.draw. Good?

Fixed Layers in svgcanvas. Canvas has history and Drawing does not. I also want to get things clean over to Layer, so I need a good, simple history service that I can inject. I came up with  HistoryRecordingService.  See the docs and let me know what you think. See Canvas.mergeLayer and mergeLayers to see how it's used. 

Hopefully you like HistoryRecordingService, I want to use it in Drawing.prototype.setCurrentLayerPosition. It's kluged right now to return {title:SVGGElement, previousName:string} because that's what the history needs. With then new mechanism, I just pass in a HistoryRecordingService.

I still need to write some tests for HistoryRecordingService, but I wanted to get your take first, before I spent the time.

flintobrien:
Just look at my earlier note "_Just saw references to all_layers in svgcanvas.js_". I should have been more clear that I had found a bug.

flintobrien:
I'm happy to do another pull request if that's the easiest way to merge again (instead of reverting the pull request [merge](https://github.com/SVG-Edit/svgedit/commit/b985a062c0389e20b410179a487150fac48723a8)). However you want to handle it.

codedread:
Reviewed the commit and I like the concept and understand why it's needed.

PR #110, opened 2016-05-04
Fixed svgCanvasToString to get latest nsMap for custom namespace support.

flintobrien:
#### Please review
I'm not sure this the best way to fix this problem? Do you have other suggestions?

This is needed to support extensions that call sanitize.addWhiteListException()
for custom namespace support. I assume this is related to [Issue #94](https://github.com/SVG-Edit/svgedit/issues/94).

svgCanvasToString was using the static nsMap from startup. This doesn't have the latest namespaces added during our extension's calls to sanitize.addWhiteListException(). Without this fix, clicking Save or SVG Source view removes all custom namespaces that were in the DOM.

#### Example Extension:
```
svgedit.editor.addExtension( "ext-fisheye-base", function ( S ) {
   svgedit.sanitize.addWhiteListException("g", 'http://www.totalgrid.org', 'some-attribute');
   ...
}
```

### DIFF
--- a/editor/svgcanvas.js
+++ b/editor/svgcanvas.js
@@ -3698,6 +3698,9 @@ var removeUnusedDefElems = this.removeUnusedDefElems = function() {
 // Returns: 
 // String containing the SVG image for output
 this.svgCanvasToString = function() {
+	// Get the latest NS after any extensions have called sanitize.addWhiteListException().
+	nsMap = svgedit.getReverseNS();
+
 	// keep calling it until there are none to remove
 	while (removeUnusedDefElems() > 0) {}
 	


flintobrien:
#### sanitize.addWhiteListException()
I just realized that [sanitize.addWhiteListException()](https://github.com/gec/svgedit/blob/fisheye/editor/sanitize.js#L105) is only in the [GEC fork, branch fisheye](https://github.com/gec/svgedit/tree/fisheye) of svgedit. I can do a pull request if others think it would be helpful.

#### NameSpace Class
If I was starting from scratch, I would use a class for namespaces that would always be internally consistent. It would have private data and accessors for names and reverse names. Doing this would change the API for svgedit.NS, svgedit.getReverseNS, etc.

It's straight forward to update the files in the project to use a new API, but what about other projects that extended svgedit. I'm not sure if updates should lean towards being backwards compatible or lean towards making the code cleaner and less entangled. It's always nice when you can do both.

#### NS Config
A simple config for adding namespaces would be nice too.

codedread:
Do you have an example of the new API?  I've forgotten how the sanitize / namespace stuff worked, but based on Issue #94 and my dim memory, we never supported custom namespaces in the official API (I remember Jacques Distler lobbying to add MathML support to it and we hard-coded support in).

This would imply that any new API introducing the ability to extend the namespaces wouldn't need to worry about being backward compatible and we just need to preserve the reading of supported namespaces.

flintobrien:
Here's an example of what I was thinking of. I haven't tested it.

Given the Namespaces class below, here's how it could be used.

```
// config.namespaces would be the same content that's currently in svgedit.NS.
svgedit.namespaces = new svgedit.Namespaces( config.namespaces);

// For backwards compatibility and convenience, we could do the following; however, if anyone
// updates svgedit.NS directly, then all bets are off. Use svgedit.namespaces.add() instead.
svgedit.NS = svgedit.namespaces.getAll();

// Also useful, but again, use svgedit.namespaces.add() to udpate.
svgedit.REVERSE_NS = svgedit.namespaces.getReverseAll();
```

### Namespaces
```
/**
 * Class for keeping track of namespaces accessible by prefix and by URI.
 *
 * NOTE: This class assumes that URIs are always unique.
 * 
 * @param {Object.<string, string>} defaultNamespaces - Object with map of prefix -> URI.
 * @constructor
 */
svgedit.Namespaces = function (defaultNamespaces) {
	this.prefixMap_ = $.extend( {},defaultNamespaces); // make a copy
	this.uriMap_ = {};
	this.updateUriMap_();
};

svgedit.Namespaces.prototype.get = function (prefix) { return this.prefixMap_[prefix]; };
svgedit.Namespaces.prototype.getAll = function () { return this.prefixMap_; };

svgedit.Namespaces.prototype.getReverse = function (uri) { return this.uriMap_[uri]; };
svgedit.Namespaces.prototype.getReverseAll = function () { return this.uriMap_; };

/**
 * Add or update a namespace
 * @param {string} prefix - namespace prefix
 * @param {string} uri - namespace URI
 */
svgedit.Namespaces.prototype.add = function (prefix, uri) {
	this.prefixMap_[prefix] = uri;
	this.updateUriMap_();
};

/**
 * Remove a namespace with the specified prefix
 * @param {string} prefix - namespace prefix
 */
svgedit.Namespaces.prototype.remove = function (prefix) {
	var uri = this.prefixMap_[prefix];
	if(uri) {
		delete this.prefixMap_[prefix];
		delete this.uriMap_[uri];
		return true;
	} else {
		return false;
	}
};

/**
 * Keep the uriMap in sync with any updates to the prefixMap.
 * Don't replace the maps, just update them so any references to the maps
 * outside this class are still valid.
 */
svgedit.Namespaces.prototype.updateUriMap_ = function () {
	var self = this;
	// delete URIs from uriMap_ that are no longer in the prefixMap_
	$.each(self.uriMap_, function( uri, prefix) {
		if( self.prefixMap_[prefix] === undefined) {
			delete self.uriMap_[uri];
		}
	});
	// Add or update namespaces in uriMap_
	$.each(self.prefixMap_, function(name, URI) {
		self.uriMap_[URI] = name.toLowerCase();
	});
};
```

codedread:
I like the idea of an internally consistent map that has an API that all code can use to get the namespace map - does it really need to be a class though (seems no point to having multiple instances of it - so maybe just an object literal with methods on it?)

One thing I see is that the getAll() method means the map is not really private data so that would be discouraged except for deprecated uses of svgedit.NS - is that the idea?

flintobrien:
Regarding "_object literal_", do you mean something like "_Refactored - Singleton via Anonymous Closure_" down below?

Regarding
> One thing I see is that the getAll() method means the map is not really private data so that would be discouraged except for deprecated uses of svgedit.NS - is that the idea?

Not bad, but not what I was thinking. The current svgedit namespace code uses a single _constant_ object, NS. This is already passing around _private_ data and no one blinks an eye. That's idiomatic Javascript, so I don't fight it. I was thinking that _getting_ a namespace could use svgedit.NS, but _modifying_ the namespaces would use the API. I also prefer to read code like ```NS.XMLNS``` rather than ```svgedit.namespaces.get('XMLNS')```.

The class was written to never replace the two internal objects, but update them. Everyone who has the internal reference is always up to date.

I don't feel a strong right or wrong here. I'd like to hear what you think.

### Details
Note: I tried to make this shorter; but, it's complicated and I think it's helpful to get on the same page, design wise - not just for namespaces.

The Javascript language and APIs often rely on API users playing nice with references to internal data, so I'm more comfortable with returning internal objects than in other languages. For example, ```this.someData_``` is  private by convention, but anyone can access it. It also helps that Javascript is single threaded and APIs are mostly non-reentrant.  You can hide data with the [Module pattern](http://www.adequatelygood.com/JavaScript-Module-Pattern-In-Depth.html); but, Javascript classes are still used extensively.

#### The Problem
What's not the problem is accessing specific namespaces (ex: NS.XML). The problem is when accessing a list of all namespaces or reverse namespaces.

The problem with svgCanvasToString (and possibly other users of namespaces) is that it retrieves the reverse namespace object at startup. The namespaces are subsequently updated, so the reverse namespaces copy is out of sync. I see three common patterns to solve this.

#### API Design Options
Which option would you choose?

1. Always retrieve the latest NS and REVERSE_NS object one-the-fly, when needed.
  * Every user could get their own copy, so there's no opportunity for mischief.
  * This may have memory and performance issues for functions that are called often.

1. Pass around mutable singletons for both NS and REVERSE_NS objects, but update via API. The objects are never replaced, only modified. This is the case with the current svgedit.NS; except, it was originally intended to be constant.
  * Since there is only one copy, they are always up to date and in sync (when managed through an API). 
  * The user doesn't have to be notified when namespaces are updated (they always have the currently updated singleton).
  * Since there's only one copy, some users could cause mischief by updating them outside of the API contract.

1. Retrieve a copy of NS and REVERSE_NS objects and get update notifications.
  * Every user could get their own copy, so there's no opportunity for mischief.
  * On update, a new copy of NS and REVERSE_NS are retrieved.
  * This is the most bullet proof solution, but it's more effort for the API code and user code.

In all cases, update notifications could be added for users that need to do something on namespace updates.

#### Usage Design Options
The current way to get the XML namespace is, for example:
```
this.svgElem_.setAttributeNS(NS.XMLNS, 'xmlns:se', NS.SE)
```
Given the _class_ below, retrieving a namespace _properly_ would be (harder to read)...
```
this.svgElem_.setAttributeNS(svgedit.namespaces.get('XMLNS'), 'xmlns:se', svgedit.namespaces.get('SE'))

```
A user (or the API) could create a helper for easier syntax:
```
function NS(prefix) { return svgedit.namespaces.get(prefix); }
// Usage would then be:
this.svgElem_.setAttributeNS(NS('XMLNS'), 'xmlns:se', NS('SE'))
```
But I'm actually fine with grabbing the internal reference:
```
var NS = svgedit.namespaces.getAll();
this.svgElem_.setAttributeNS(NS.XMLNS, 'xmlns:se', NS.SE)
```

---
<br/>

### Refactored - Singleton via Anonymous Closure
```
(function(svgedit, $, defaultNamespaces) {
'use strict';

var prefixMap = $.extend( {},defaultNamespaces); // make a copy
var uriMap = {};

/**
 * Keep the uriMap in sync with any updates to the prefixMap.
 * Don't replace the maps, just update them so any references to the maps
 * outside this class are still valid.
 */
function updateUriMap() {
    // delete URIs from uriMap that are no longer in the prefixMap
    $.each(uriMap, function(uri, prefix) {
        if( prefixMap[prefix] === undefined) {
            delete uriMap[uri];
        }
    });
    // Add or update namespaces in uriMap
    $.each(prefixMap, function(name, URI) {
        uriMap[URI] = name;
    });
};
updateUriMap();

var exports = {};

exports.get = function (prefix) { return prefixMap[prefix]; };
exports.getAll = function () { return prefixMap; };

exports.getReverse = function (uri) { return uriMap[uri]; };
exports.getReverseAll = function () { return uriMap; };

/**
 * Add or update a namespace
 * @param {string} prefix - namespace prefix
 * @param {string} uri - namespace URI
 */
exports.add = function (prefix, uri) {
    prefixMap[prefix] = uri;
    updateUriMap();
};

/**
 * Remove a namespace with the specified prefix
 * @param {string} prefix - namespace prefix
 */
exports.remove = function (prefix) {
    var uri = prefixMap[prefix];
    if(uri) {
        delete prefixMap[prefix];
        delete uriMap[uri];
        return true;
    } else {
        return false;
    }
};

svgedit.namespaces = exports;

}(svgedit, $, config.defaultNamespaces));
```

codedread:
Gotcha.  I didn't have time to think about this too deeply, but I think I'd vote for "Pass around mutable singletons for both NS and REVERSE_NS objects, but update via API"

brettz9:
@flintobrien : A PR for `addWhiteListException` SGTM... as with any new or updated PR for the namespaces...

jfhenon:

PR #132, opened 2016-08-12
Updated/reviewed French translation

Frigory33:
Hi,
I brought some corrections to the French translation and translated some strings which were still in English.
I'm also a translator for the Inkscape project — but not really skilled at vector graphics actually, more at translation.
Thanks for merging and congratulations for your project.


MrRio:
Thanks very much!

Issue #52, opened 2015-12-22
update jquery (and jqueryui) -- 1.7.1 is 4 years old


sebhaase:
SVG-Edit currently uses jQuery 1.7

version 1.8 or better 1.9 would be start towards jQuery 2.x, which is current.

E.g.  jQuery.andSelf  could be replaced with jQuery.addBack

GoTeamScotch:
I agree with this. On that note, I've been using jQuery v1.11.3 in my local devleopment copy with no apparent problems resulting from using the newer version. I haven't gone through the unit tests though so there could be some complications present.

sivasaravanand:
please let me know what are the files needs to change in order to reference latest jquery file. i tried but throwing many script errors

Harvie:
also please everywhere load the jquery using HTTPS!!!!

brettz9:
I'm working on a PR to avoid CDN use entirely (applies to imagelib and mathjax) (the latter URL wasn't working). This should help with local testing/offline apps as well as ensure future compatibility. Optional use of CDN could be added (PRs welcome).

There is a "jquery_release" variable in an HTML comment which was apparently intended to conditionally load the CDN version, but I'm not sure what build process uses that, as I couldn't find that elsewhere in the repo. Ideally any update to these older versions that I've copied will be added through npm instead, ensuring a smoother upgrade path.

brettz9:
I've at least updated the even older version of jQuery used by the Imagelib extension (it had been using a 1.4.4 copy)...

One approach I've been considering (and I'd like your feedback) is to actually move away from jQuery in places if not entirely, using ES6, DOM enhancements (like ChildNode and ParentNode interfaces), and focused libraries, such as my project [Jamilih](https://github.com/brettz9/jamilih/)--which aims at a straightforward, light-weight approach to building the DOM--there is less need for jQuery.

And no doubt there are some light-weight, single-purpose substitute micro-libraries for traversal, etc. as well (https://github.com/WebReflection/query-result looks interesting to me despite its age). 

There are advantages to jQuery and frameworks, but with the magic can come less focused and breaking version changes, and some bloat, and, as we're experiencing now, difficulty to migrate.

What do people think?

jfhenon:
This issue has no activity for more than 2 years.

sebhaase:
Please don’t close tickets just because they are old.

PR #135, opened 2016-08-25
make most functions work locally without a Node server (enhancement), i18n fixes, getScript failure recovery (bug)

duzc2:
Hello .  I am trying to make functions work locally without a server.
it works great now , but just lose some svg icons.


editor/extensions/ext-markers.js
```diff
@@ -50,7 +50,7 @@ svgEditor.addExtension("Markers", function(S) {
 	var marker_types = {
 		nomarker: {},  
 		leftarrow:  
-			{element:'path', attr:{d:'M0,50 L100,90 L70,50 L100,10 Z'}},
+			{element:'path', attr:{d:'M0,50 L100,90 L70,50 L100,10 Z'},icon:svgEditor.curConfig.extPath + 'markers-icons-leftarrow.png'},
```
codedread: what is this about?

brettz9:
I wonder if `$.ajax` in later jQuery versions allows for same-directory `file://` retrieval (perhaps with `isLocal`), so that this would be a non-issue if our jQuery version were upgraded as per #52 .... In the meantime (or instead), I think a better PR might simply replace our current use of `$.ajax` (once in `svg-editor.js` and once in `jQuery.svgIcons.js`) with our own `XMLHttpRequest`; in that way, the SVG icons should work locally without need for fallbacks to PNGs.

brettz9:
Nevermind, it appears Chrome just won't allow `file:` locally for XHR.

brettz9:
I've incorporated your fixes (i18n (which I confirmed) and images (which I expanded upon)). The images have also been optimized, and I've confirmed they are being applied locally when run from `file://`.

PR #145, opened 2016-11-02
Added a missing se:connector attribute to the polyline tag on the whitelist

iuyiuy:
The se:connector attribute was missing in the whitelist for the polyline tag, making saved connector lines not work when loading the drawing after.


codedread:
It seems like this PR is not necessary, since I already have a separate section dealing with se: attributes.  Search for the comment "Programmatically add the se: attributes"

iuyiuy:
Sorry, this is because i was running a branch that included the changes in the issue_94 branch, where i removed that handling in order to set svgedit.ignoredNS to namespaces that will not be filtered by the whitelist. I've now made the change in the other branch. You can close this PR.

PR #147, opened 2016-11-08
Fixed issue with having multiple includeWith stanzas in different extensions

iuyiuy:


codedread:
Sorry I didn't understand this change - can you clarify a little more what it does?

iuyiuy:
In extensions/ext-connector.js there's a button specified with an includeWith object, which defines that it should be included in the same flyout menu as the line tool. When i tried to make an extension, which included another button in the same menu, it didn't work. These changes made that work.

editor/svg-editor.js
```diff
@@ -2037,20 +2041,6 @@ TODOS
...
-									$.each(holders[opts.parent], function(i, tool) {
-										if (tool.sel == currentOperation) {
-											if (!event.shiftKey || !flyoutIsSelected) {
-												options = tool;
-											} else {
-												options = holders[opts.parent][i+1] || holders[opts.parent][0];
-											}
-										}
-									});
-								}
```
codedread: Why is this code being removed?

editor/svg-editor.js
```diff
@@ -2019,14 +2019,18 @@ TODOS
...
 					var shower = $(show_sel);
 					var def = false;
-					buttons.addClass('tool_button')
+					buttons.addClass('tool_button tool_button_evt_handled')
 						.unbind('click mousedown mouseup') // may not be necessary
-						.each(function(i) {
+						.each(function() {
 							// Get this buttons options
+							var i;
+							for(var j in btn_opts)
```
codedread: Also, you should break after i = j, correct?

editor/svg-editor.js
iuyiuy: My commit message for that change said "Removed some code that would select the wrong tool when using keyboard shortcuts for flyout buttons." On the svg-edit demo page you might notice that the keyboard shortcut for the line tool [L] doesn't work properly:
https://svg-edit.github.io/svgedit/releases/svg-edit-2.8.1/svg-editor.html

I haven't gone back and tested it now, but i believe this code was part of the issue and didn't seem to serve a purpose. I haven't been able to find a keyboard shortcut that doesn't work as expected after this change.

editor/svg-editor.js
iuyiuy: Of course. Thanks for spotting that. These changes are now committed and pushed.

editor/svg-editor.js
codedread: Ok, I played around with it on the trunk and it seems that pressing L selects the line tool but it doesn't show up in the toolbar as selected, while pressing R selects the rectangle tool and it shows as selected.  I guess you are reasonably sure this doesn't break anything so let's go for it.

editor/svg-editor.js
```diff
@@ -2029,8 +2029,11 @@ TODOS
 						.each(function() {
 							// Get this buttons options
 							var i;
-							for(var j in btn_opts)
-								if(btn_opts[j].sel == '#'+this.getAttribute('id')) i = j;
+							for(var j in btn_opts) {
+								if(btn_opts[j].sel == '#'+this.getAttribute('id')) {
```
codedread: And a space after "for" and before (

brettz9:
Fix applied in `master` just now (https://github.com/brettz9/svgedit/commit/0a2da503f1136bff67bf884fd01d1cf4abae0800) which incorporates this PR. I included the `keypress` code portion, however, albeit with adjustments, as that was serving the purpose of allowing `shift + <key>` to cycle through the flyouts. Closing therefore but feel free to comment further.

PR #152, opened 2016-11-29
SVG validity check in svgCanvas.setSvgString()

iuyiuy:
Added a check for valid svg (or at least the namespace), so we can return false right away, instead of relying on it to be caught in a try/catch with a not so descriptive error message.

### DIFF
--- a/editor/svgcanvas.js
+++ b/editor/svgcanvas.js
@@ -4446,6 +4446,8 @@ this.setSvgString = function(xmlString) {
 	try {
 		// convert string into XML document
 		var newDoc = svgedit.utilities.text2xml(xmlString);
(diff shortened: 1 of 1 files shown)


editor/svgcanvas.js
```diff
@@ -4446,6 +4446,8 @@ this.setSvgString = function(xmlString) {
 	try {
 		// convert string into XML document
 		var newDoc = svgedit.utilities.text2xml(xmlString);
+		if(newDoc.firstChild.namespaceURI != 'http://www.w3.org/2000/svg')
+			return false;
```
brettz9: `firstChild` should probably be `firstElementChild`, or rather, use jQuery for backward-compatibility, to ensure it is referencing the first element and not erring by trying to access a namespace on a comment node or processing instruction.

PR #155, opened 2016-12-02
Added ability to set namespace in addSvgElementFromJson().

iuyiuy:


brettz9:
I'd think the method name should be renamed (to omit "SVG") if it will be adding an element which may not be SVG...

editor/svgcanvas.js
```diff
@@ -200,7 +200,8 @@ var addSvgElementFromJson = this.addSvgElementFromJson = function(data) {
 		shape = null;
 	}
 	if (!shape) {
-		shape = svgdoc.createElementNS(NS.SVG, data.element);
+		var ns = data.namespace ? data.namespace : NS.SVG;
```
brettz9: This can also be simplified to `var ns = data.namespace || NS.SVG;`

PR #168, opened 2017-01-24
Bind extensions click events as mousedown (touch devices)

ClemArt:
On touch devices the click event is not used, instead touch.js fallbacks
to mousedown.

Thus extension buttons listening to "click" could not be selected nor extension tools used (as
panning, color picking, star shape) on a touch device (or even touch emulation in google chrome).

If svg recognize a touch device, click events for extensions are now bind as
mousedown.


brettz9:
I reworded the comment and applied on top of our latest code, now in `master`. Thanks for the clearly described explanation as well as the PR!

PR #169, opened 2017-02-01
Fix zooming so that it centers on the cursor

AndrolGenhald:


GoTeamScotch:
Kind of unrelated but here's a slight tweak that only enables zoom on scroll if you have the Zoom tool enabled. Press Z to enter zoom mode and V to exit. I find myself wanting to scroll the canvas more often than adjusting the zoom. With this you get the best of both worlds.

`if (svgCanvas.getMode() !== 'zoom') { //monkey
	return;
}`

brettz9:
I've incorporated this very needed fix, thanks very much! (Adapted to follow some additional conventions...)

Issue #175, opened 2017-04-09
Faster Freehand onMouseMove?


jtnw:
I'm integrating SVGEdit into an application and with all the other javascript the freehand draw is quite slow. The slowdown is more noticeable in my application, but it does happen in the official editor. To reproduce, use the freehand tool and keep drawing a single path until you notice the frame drop. Mainly, I've made this faster by lowering the STEP_COUNT constant, but am also looking at other ways to speed it up. 

On line 1835 of svgcanvas.js I replaced `shape.setAttributeNS(null, 'points', d_attr);` with:

    var point = svgcontent.createSVGPoint();                                       
    point.x = bSpline.x;                                                           
    point.y = bSpline.y;                                                           
    shape.points.appendItem(point);            

Although I'm not positive it makes a difference. Can someone smarter than me run a test or such to determine if this should be used instead of rewriting the whole points attribute? The theory is that the browser doesn't have to re-parse the whole points string.

jtnw:
I just tested on chrome and it does seem like it's a huge improvement. Not sure if it's an improvement on Firefox. Still I'd like to see an official test by someone if possible!

codedread:
Feel free to send a pull request :)

jtnw:
pull request sent to master branch.

jtnw:
Posting here that Robertcck found the pull request to improve speed on IE and Edge.

brettz9:
Confirmed in Firefox and Chrome to be improving performance by about two orders of magnitude (Since mousemove can't be so readily simulated, I wrapped the relevant portion within a loop and logged results when drawing manually). Thanks for the issue and PR!

PR #176, opened 2017-04-09
changed fhpath onMouseMove code to use a hopefully faster method

jtnw:
This pull request changes how the path is drawn during onMouseMove while using the freehand path tool. Instead of adding the points to the d_attr string the native SVG method is used to add the points to the polyline. Hopefully this can improve the performance of the freehand path tool which has noticeable framerate issues when drawing a very long path. If this pull request makes it to master branch, we can also remove the d_attr variable.

I've only tested on Firefox 52 and Chromium 57 on Linux. It seems faster on Chromium, but I don't have hard numbers to show it. Theoretically it should be faster though.


brettz9:
Cross-linking to issue this is meant to resolve: #176

brettz9:
Closing as integrated per https://github.com/SVG-Edit/svgedit/issues/175#issuecomment-407998401 . Thanks for the PR!

PR #188, opened 2017-08-03
Fix documentation markdown

bronzehedwick:
Github changed it's markdown parser, requiring a space between the hash and the text for headings.


MrRio:
Thanks!

PR #201, opened 2017-11-27
Fix broken saving of svg in Chrome >=60

### LINKED ISSUE #193: Chrome/Opera browser not save, export.
I try to export or save image but not work .
The result of this is a redirect to a new tab web page : about:blank

TeemuKoivisto:
Fixes #193 which was due to https://stackoverflow.com/questions/45603201/window-is-not-allowed-to-navigate-top-frame-navigations-to-data-urls

I don't know if there is something else also broken but this fixes my use case so I don't have a lot of motivation to spend more time on this 😃

### DIFF
--- a/editor/svg-editor.js
+++ b/editor/svg-editor.js
@@ -1049,8 +1049,13 @@ TODOS
 					return;
 				}
 
-				// Opens the SVG in new window
-				var win = wind.open('data:image/svg+xml;base64,' + Utils.encode64(svg));
+				// Since saving SVGs by opening a new window was removed in Chrome use artificial link-click
+				// https://stackoverflow.com/questions/45603201/window-is-not-allowed-to-navigate-top-frame-navigations-to-data-urls
+				var a  = document.createElement('a');
+				a.href = 'data:image/svg+xml;base64,' + Utils.encode64(svg);
+				a.download = 'icon.svg';
+		
+				a.click();
 
 				// Alert will only appear the first time saved OR the first time the bug is encountered
 				var done = $.pref('save_notice_done');


editor/svg-editor.js
```diff
@@ -3762,6 +3767,7 @@ TODOS
 			};
 
 			var saveSourceEditor = function() {
+				debugger
```
codedread: Remove

editor/svg-editor.js
```diff
@@ -1049,8 +1049,13 @@ TODOS
...
 
-				// Opens the SVG in new window
-				var win = wind.open('data:image/svg+xml;base64,' + Utils.encode64(svg));
+				// Since saving SVGs by opening a new window was removed in Chrome use artificial link-click
+				// https://stackoverflow.com/questions/45603201/window-is-not-allowed-to-navigate-top-frame-navigations-to-data-urls
+				var a  = document.createElement('a');
+				a.href = 'data:image/svg+xml;base64,' + Utils.encode64(svg);
+				a.download = 'icon.svg';
+		
+				a.click()
```
codedread: Please end in a semicolon

TeemuKoivisto:
@codedread Done

editor/svg-editor.js
```diff
@@ -1049,8 +1049,13 @@ TODOS
 					return;
 				}
 
-				// Opens the SVG in new window
-				var win = wind.open('data:image/svg+xml;base64,' + Utils.encode64(svg));
+				// Since saving SVGs by opening a new window was removed in Chrome use artificial link-click
+				// https://stackoverflow.com/questions/45603201/window-is-not-allowed-to-navigate-top-frame-navigations-to-data-urls
+				var a  = document.createElement('a');
+				a.href = 'data:image/svg+xml;base64,' + Utils.encode64(svg);
+				a.download = 'icon.svg';
```
codedread: Ideally this should use the title of the document (.svg), but I think this is good enough for now.

PR #206, opened 2018-01-18
Implemented a clipboard that works across tabs and windows.

iuyiuy:
With this commit it's now possible to cut'n'paste svg elements across tabs and/or windows. It uses localStorage to achieve this.

### DIFF
--- a/editor/extensions/ext-connector.js
+++ b/editor/extensions/ext-connector.js
@@ -160,6 +160,7 @@ svgEditor.addExtension("Connector", function(S) {
 
 		// Loop through connectors to see if one is connected to the element
 		connectors.each(function() {
+			var connector = this;
 			var add_this;
 			function add () {
 				if ($.inArray(this, elems) !== -1) {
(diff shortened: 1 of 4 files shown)


editor/extensions/ext-connector.js
```diff
@@ -160,19 +160,32 @@ svgEditor.addExtension("Connector", function(S) {
 
 		// Loop through connectors to see if one is connected to the element
 		connectors.each(function() {
+			var connector = this;
```
codedread: nit: can use .bind(this) on the function below instead of this

editor/extensions/ext-connector.js
```diff
@@ -160,19 +160,32 @@ svgEditor.addExtension("Connector", function(S) {
...
 			}
-			var start = elData(this, "c_start");
-			var end = elData(this, "c_end");
-			
-			var parts = [getElem(start), getElem(end)];
+
+			// Grab the ends
+			var parts = [];
+			['start', 'end'].forEach(function (pos, i) {
+				var part = elData(this, 'c_'+pos);
```
codedread: nit: use a variable for the key:

  var key = 'c_' + pos;

editor/extensions/ext-connector.js
```diff
@@ -580,6 +592,18 @@ svgEditor.addExtension("Connector", function(S) {
 				updateConnectors();
 			}
 		},
+		IDsUpdated: function(input) {
+			var remove = [];
+			input.elems.forEach(function(elem){
+				if('se:connector' in elem.attr) {
+					elem.attr['se:connector'] = elem.attr['se:connector'].split(' ')
+						.map(function(oldID){ return input.changes[oldID] }).join(' ');
+					if(!/. ./.test(elem.attr['se:connector']))
```
codedread: Can you add a comment here, it's not clear what this regex test is doing

editor/svgcanvas.js
```diff
@@ -177,6 +177,30 @@ var cur_shape = all_properties.shape;
...
+    if(data.nodeType == 3) return data.nodeValue;
+
+    var retval = {
+        element: data.tagName,
+        //namespace: nsMap[data.namespaceURI],
+        attr: {},
+        children: [],
+    };
+
+    // Itrate attributes
```
codedread: nit: Iterate

iuyiuy:
I've implemented the changes you mentioned.

PR #207, opened 2018-01-19
Separate code and configuration

iuyiuy:
I need to separate the code and configuration, so i can have them in separate repositories, since it's really useful to manage them separately. This is my suggestion for how to do it. It should be generically useful.


iuyiuy:
The point of it is so we can have the configuration and the CSS file outside the repository. This way the folder with the svg-edit code (the editor folder) can be versioned separately. The files weren't in the repo before either. But with the files being outside the editor folder, we can have it point to files that are in a different repo and therefore managed separately.

codedread:
I understand now - sorry for the confusion.  Can you please update the instructions in config-sample.js ?

codedread:
Also, won't this cause error messages in the dev console if the files don't exist?

iuyiuy:
I've updated the instructions.

It does cause error messages, but the file wasn't there already, so that wasn't introduced by these commits. You can even see it in the demo:
http://www.clker.com/inc/svgedit/svg-editor.html

I'm not sure if we should address that and how, but i'd suggest handling that in a new bug report.

brettz9:
This is a good idea, but as I've published us to npm, we may want to default to three directories up (out of `editor`, out of `svgedit` and out of `node_modules`).

Also, with a move I'm working on now toward ES6 modules, we might also want an alternative HTML file which indicates the config file as a module and lets the user's config file import SVG-edit (for anyone directly using ES6 modules in production as is now becoming possible in modern browsers). (I am also planning on asynchronously loading the stylesheets though blocking until all loaded, mimicking HTML behavior, giving us modularity (though with performance) with the CSS as well.)

While I suspect SVG-edit may require the full page for proper styling, I wonder if it is possible for us to avoid the need for an iframe to provide the embedded API, and just let users load our library which will add all of the HTML dynamically, even into a container other than `<body>`.

brettz9:
I've incorporated the changes for now until we may reconsider a packaging approach. Thanks!

PR #208, opened 2018-01-22
Add ability to set the svg drawing without adding to the undo stack.

iuyiuy:
When loading a file on init, it's useful to set it without it being added to the undo stack.

### DIFF
--- a/editor/svgcanvas.js
+++ b/editor/svgcanvas.js
@@ -3817,7 +3817,7 @@ this.svgToString = function(elem, indent) {
 				var el = this;
 				// for some elements have no attribute
 				var uri = this.namespaceURI;
-				if(uri && !nsuris[uri] && nsMap[uri] && nsMap[uri] !== 'xmlns' && nsMap[uri] !== 'xml' ) {
+				if (uri && !nsuris[uri] && nsMap[uri] && nsMap[uri] !== 'xmlns' && nsMap[uri] !== 'xml' ) {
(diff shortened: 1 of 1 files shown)


editor/svgcanvas.js
```diff
@@ -4596,7 +4596,7 @@ this.setSvgString = function(xmlString) {
 		svgedit.path.clearData();
 		svgroot.appendChild(selectorManager.selectorParentGroup);
 		
-		addCommandToHistory(batchCmd);
+		if(!preventUndo) addCommandToHistory(batchCmd);
```
codedread: nit: add a space between the if and the (

also, can you please correct the other two instances of that in this file?

iuyiuy: I only found one other instance of it in this branch, but i've changed that now.

I might have done it in that style in other commits / pull requests. Sorry about that. I didn't realize that was the coding style of this project. I'll try to do that in the future.

editor/svgcanvas.js
```diff
@@ -4442,7 +4442,7 @@ var convertToGroup = this.convertToGroup = function(elem) {
 //
 // Returns:
 // This function returns false if the set was unsuccessful, true otherwise.
```
codedread: Please update the docs of this method to mention the optional preventUndo parameter

iuyiuy: I've added the documentation for it now.

PR #215, opened 2018-02-16
Removed BOM encoding and lossless image compression

drsassafras:
All files have identical content with the exception of the UTF-BOM encoding flag removed. (can cause weird bitewise errors if not handled correctly) and a reduction in file size for images. Every pixel is identical. Google requires lossless image compression for best search results.


brettz9:
I think the latest commits have removed the BOM marks, so you might want to rebase this PR accordingly.

brettz9:
Also, it'd be nice to know how to replicate how you're doing the compressions...

drsassafras:
ImageOptim is good software and easy to use.
FileOptimizer achieves better compression but is less reliable and not as user friendly.

I always check to make sure each image is exactly the same afterwards as well.

drsassafras:
I re-checked for BOM files. Seems you missed 8 of them.

svgedit-bom

drsassafras:
#237 

Created another pull request for the BOM encoding.

brettz9:
Thanks very much for the PRs--I'm working on a PR to apply image compression and the BOM along with the other changes I'm working on.

brettz9:
These improvements have now been incorporated in `master`--thanks very much!

Issue #216, opened 2018-02-27
Editing paths does not work in Firefox 59+


progers:
Firefox 59 removed some SVG path seg APIs:
https://bugzilla.mozilla.org/show_bug.cgi?id=1436438

This breaks path editing in svgedit which now throws:
"TypeError: u.clear is not a function"

I've updated [pathseg.js](https://github.com/progers/pathseg) to fix this issue, we just need to import the latest version into svgedit. I checked that Firefox 58, Firefox 59, Chrome 65, and Safari Tech Preview can all edit paths correctly with svg edit and the latest version of pathseg.js.

progers:
Fixed!

PR #217, opened 2018-02-27
Update pathseg.js to fix Firefox 59+ path editing

progers:
Firefox 59 removed some SVG path seg APIs which breaks path editing.

This change updates pathseg.js to the latest version which contains a fix for Firefox 59. I checked that Firefox 58 (pre-api-change), Firefox 59 (post-api-change), Chrome 65, and Safari Tech Preview can all edit paths correctly with this change. This change has a lot of diffs due to minor changes in pathseg.js since the first time it was imported into svgedit.

This fixes issue #216


codedread:
Thanks!  I guess this means we need to do a release to make sure the trunk version isn't broken :(

progers:
@codedread, yeah, we should probably create another release. Lets wait a couple weeks to make sure these changes can bake, and make sure firefox goes ahead with this change. If you can remind me in a couple weeks, I can put together a release.

PR #228, opened 2018-05-09
Update not supported page

NeilFraser:
Update old links.  Remove now-defunct link to Chrome Frame.  Use CSS instead of JavaScript and jQuery.


NeilFraser:
Screwed up the branches.  Deleting and resubmitting...

PR #236, opened 2018-05-17
Lint (v2)

NeilFraser:
Non-colliding PR for remaining lint issues, as well as the zero-width circle Chrome issue.


brettz9:
Great work, thanks!

PR #237, opened 2018-05-18
Remove BOM encoding

drsassafras:


brettz9:
I've removed the BOM in these files in `master`. Thanks for the PR!  (I plan to add the image optimization in another commit)

drsassafras:
Much appreciated. We're always looking for a smaller footprint in Tiki, and now that the BOM files are gone, there is no chance of us pointing our finger in this direction the next time we have some strange bitwise errors. :)

Issue #240, opened 2018-05-24
Migrate to JSDoc (along with package.json script)


brettz9:
[JSDoc](http://usejsdoc.org/) supports [es2015-classes](http://usejsdoc.org/howto-es2015-classes.html), etc.

tetedacier:
Hello, I provided basic documentation generation update in a [PR](https://github.com/SVG-Edit/svgedit/pull/244).

brettz9:
Sweet! Thanks for your help on this!

brettz9:
See #244 on issue with exports...

brettz9:
FYI, I've put the jsdoc output within `/docs/jsdoc/` (though hidden in `.gitignore`), published it on `gh-pages` (without the `.gitignore`) (and linked to this from our `master`'s README): https://svg-edit.github.io/svgedit/releases/svg-edit-3.0.0-alpha.4/docs/jsdoc/index.html . Thanks again!

PR #244, opened 2018-05-29
Feature/jsdoc

tetedacier:
Solve #240

### DIFF
--- a/.jsdoc/.gitignore
+++ b/.jsdoc/.gitignore
@@ -1,0 +1,2 @@
+# jsdoc's output folder
+out
--- a/.jsdoc/configuration.json
+++ b/.jsdoc/configuration.json
@@ -1,0 +1,21 @@
+{
+  "recurseDepth": 10,
+  "source": {
+    "exclude": [
+      "node_modules",
+      "dist"
+    ]
+  },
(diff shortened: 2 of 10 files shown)


brettz9:
Very handy, thanks for getting this going! I think I found the rest of the ones that needed conversion to JSDoc, and it is passing... 

One issue if you happen to know how to fix--a file is getting generated within the docs (at least on my Mac), `module.html#.exports`, with the hash part of the file name (thus not working as a proper HTML file)...

tetedacier:
I will give a look at it and submit a PR with most of document cleaned.
I known this PR was just the bootstrap of something bigger 😉

tetedacier:
After some test this file seems to be a catch all for function/class defined directly in the export section.

The best solution to avoid this issue is to declare them outside of the export section. Note that at least [one bug of the jsdoc issues](https://github.com/jsdoc3/jsdoc/issues/1517) is related to this.

If this solution is ok for you, I will update those place where definition are done in export section.

brettz9:
Ok, I think that's a pretty small price to pay for working docs... That'd be nice, thanks!

tetedacier:
You're welcome, I'm on it.

PR #245, opened 2018-05-29
Bugfix/separate declaration from exposition

tetedacier:
This PR separate declaration of js item and their exposition. It fixes issue with jsDoc when both are mixed.

This PR follow the discussion we had with @brettz9 on [jsdoc PR](https://github.com/SVG-Edit/svgedit/pull/244)

Moreover, I increase versosity of the tasks and suppress previously generated documentation before generating new one.


brettz9:
Excellent--nice work!

PR #253, opened 2018-06-05
Importing issues/changes

### LINKED ISSUE #246: Regression: Document Properties styling issue
There is a problem, probably a regression, with the `SVG-Edit` -> `Document Properties` dialog, as it is now showing up malpositioned and partly transparent.

It'd be great if someone more skilled with CSS could take a look, even just to identify what the styling should be, even if not identifying the responsible misplaced CSS or JavaScript setting of the CSS.

You can see it on `master` or on the last alpha release: https://svg-edit.github.io/svgedit/releases/svg-edit-3.0.0-alpha.3/editor/svg-editor-es.html

### LINKED ISSUE #249: Regression with PDF export
Some recent changes have caused errors during PDF export. Need to fix.

brettz9:
A bunch of changes, some to fix the known remaining regressions, and some in the spirit of the release (e.g., ES6 refactoring, having locales and extensions formatted as simple ES6 module exports rather than requiring programmatic calls, getting in other minor breaking changes for next major version)

- Breaking change: Extension now formatted as export (and `this` is set to editor, including for `callback`)
- Breaking change: Locale now formatted as export
- Breaking change: Moved out remaining modular i18n (imagelib) to own folder
- Breaking change: Drop `executeAfterLoads` (and getJSPDF/getCanvg)
- Breaking change: `RGBColor` must accept `new`
- Breaking change: canvg - `stackBlurCanvasRGBA` must be set now by function (`setStackBlurCanvasRGBA`) rather than global; `canvg` now a named export
- Breaking change: Avoid passing `canvg`/`buildCanvgCallback` to extensions (have them import)
- Fix (regression): Add polyfill for `ChildNode`/`ParentNode` (and use further)
- Fix (regression): Positioning of Document Properties dialog (Fixes #246)
- Fix (regression): PDF Export (Fixes #249)
- Fix (regression): Apply Babel universally to dependencies
- Fix (regression): Ordering of `uaPrefix` function in `svgEditor.js`
- Fix (regression): Embedded API
- Fix (embedded editor): Fix backspace key in Firefox so it doesn't navigate out of frame
- Fix: i18nize imaglib more deeply
- Fix: Alert if no exportWindow for PDF (e.g., if blocked)
- Refactoring( RGBColor) `RGBColor` as class, without rebuilding constants, optimize string replacement, move methods to prototype, use templates and object literals, use `Object.keys`
- Refactoring (canvg) Use classes more internally, use shorthand objects; array extras, return to lazy-loading
- Refactoring: Use Promises in place of `$.getScript`; always return Promises in case deciding to await resolving
- Refactoring: Avoid importing `RGBColor` into `svgutils.js` (jsPDF imports it itself)
- Refactoring: Arrow functions, destructuring, shorter property references
- Refactoring: Fix `lang` and `dir` for locales (though not in use currently anyways)
- Refactoring: Provide path config for canvg, jspdf


brettz9:
Great, thanks! :-)

PR #271, opened 2018-08-08
Add issue templates

brettz9:


brettz9:
@codedread : If you have any additional thoughts or ideas on what to add, thought we'd make use of this issue template system given how frequently we need to know at least which version people are using...

Issue #276, opened 2018-08-22
nested svg element does not move after mousemove event


initialdmg:
[svgcanvas.js](https://github.com/SVG-Edit/svgedit/blob/master/editor/svgcanvas.js#L2069) implements the selected element's move event, it works well on pure svg element, like `<rect>`, `<text>`...
i writed a extension, it's icon has svg elements nested in the outer svg icon, like

```html
<svg >
  <g class="layer">
    ...
    <g id='extension' />
      <rect></rect>
      ......
      <svg >
        ...
      </svg>
    <g>
  </g>
</svg>
```
everything is fine when draging, while, after mouseup event, the nested element back to its initial location. Nested SVG element's x and y attributes stay their initial value.  I tried the setTranslate() method on the root svg node, it works correctly.

initialdmg:
Hi, @brettz9!  I've checked PR 159 you mentioned, it's not what i mean. I've create a PR to resolve this. : )

brettz9:
Great! Could you prepare a sample file to reproduce? That should facilitate review...

initialdmg:
I wrote an extension , `extensions\ext-test.js`, to reproduce the bug. I find that version 3 on master branch seems to have resolved the bug :joy: :+1:, see: [https://rawgit.com/initialdmg/svgedit/fix-nested-svg/editor/svg-editor.html](https://rawgit.com/initialdmg/svgedit/fix-nested-svg/editor/svg-editor.html)

While it replays on the version 2.8.1 which i'm using, see: [https://rawgit.com/initialdmg/svgedit/replay-issue-2.8.1/editor/svg-editor.html](https://rawgit.com/initialdmg/svgedit/replay-issue-2.8.1/editor/svg-editor.html)

brettz9:
Hmm... I'm not sure I am clear on how to reproduce this, and things seem to behave similarly to me in both versions. But if the bug is resolved, I'll go ahead and close the issue and PR, though feel free to comment further or reopen if there are issues remaining.

PR #277, opened 2018-09-04
 fix: nested svg element do not transform

initialdmg:
fix for issue #276


brettz9:
Per #276 , it sounds like the issue was invalid per version 3 changes, so closing. Feel free to reopen otherwise.

PR #307, opened 2018-12-04
Fix Grid bug

NeiroNx:
Fix grid pattern creation. Created in `svgroot` but needs only in `canvasGrid` element. Move it into `canvasGrid` `defs` element to prevent conflict with markers.

### DIFF
--- a/dist/extensions/ext-grid.js
+++ b/dist/extensions/ext-grid.js
@@ -51,7 +51,7 @@ var svgEditorExtension_grid = (function () {
       var _init = _asyncToGenerator(
       /*#__PURE__*/
       regeneratorRuntime.mark(function _callee(_ref) {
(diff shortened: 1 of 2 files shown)


brettz9:
Awesome, thanks for the fix...

FYI, I've updated `master` to fix some eslint issues which may have prevented `npm run browser-test` (which should be run to handle the builds). I also updated Babel and updated the build files accordingly...

Thanks again!

NeiroNx:
Now i am checking markers and try to fix some bugs. 
And extend it functionality: change size of marker directly from line strokewidth. 
And as expected - i need additional size in pannel - may be need two line pannel?
first line - actions with object,
second line - params of object

brettz9:
It seems the top panel has some extra space... One might also have expanders as with the existing line stroke width on the bottom... Not sure if others have opinions, but otherwise I'd suggest just going with what you'd like to use...

PR #309, opened 2018-12-07
Add russian translation to some strings

NeiroNx:


brettz9:
Thanks for this!

FYI, there are also multiple locales within `editor/extensions/ext-locale` for extension localization (with extension localization now defined modularly).

Issue #320, opened 2019-01-24
Dimensions passed on URL don't work


sjernigan:
The doc mentions that you can pass image dimensions on the URL like so
`.../svg-editor.html?dimensions=300,240&canvas_expansion=5&initFill[color]=0000FF`

However, when I try to do that with master or several hosted versions, I get a
`TypeError: arr[Symbol.iterator] is not a function`
error in the console.  I could not find a workaround format.  

I believe I understand the current status of this project so many thanks for any responses.

sjernigan:
Adding the below just above the call to config_.dimensions solves the issue for me but I may not understand the issue.  I can submit a PR if desired.
`
    if (config_.dimensions[Symbol.iterator] != 'function') {
      config_.dimensions = [config_.dimensions[0],config_.dimensions[1]]
    }
`

brettz9:
Ok, there was one issue I've fixed just now in d7b9cc630950e50e1f81608fe9e98bd5d4aa45bb that array configs were not being preserved internally as arrays. You'll have to see though what other issues might still be present though as far as dimensions--e.g., updating the Document Properties dialog, etc.

sjernigan:
The changes seem to work.  I have not noticed other issues though I only use dimensions, url, lang, and storagePrompt.  Thank you.

brettz9:
Are the dimensions to be reflected in the Document Properties dialog?

sjernigan:
Yes, they are correct in the Document Properties

PR #321, opened 2019-01-24
Ref Issue #320

sjernigan:
#320


sjernigan:
Hold up...Found that 'new image' needs the same treatment.

sjernigan:
Well, 'New Image' is working but I think I am doing this in the wrong place.  This transformation should be made when first reading the url params.  I'm not sure where that happens yet.

sjernigan:
brettz9 has provided a more complete fix referenced from the issue.

PR #330, opened 2019-03-07
Add some examlpe

NeiroNx:
That example created with SVG-Edit with redrawing a picture found in internet.


brettz9:
Those likenesses are trademarked, so I'm afraid we'd need to refrain from redistributing. Closing, but feel free to comment further.

PR #348, opened 2019-07-15
Fixed input width

seahindeniz:
2.2em is too low for long numbers and it needs to be increased

image
<sub>After < Before<sup>


brettz9:
Nice fix, thanks!

PR #349, opened 2019-08-02
Lock axis on Ctrl

### LINKED ISSUE #346: Allow resize on a single axis with Ctrl+drag
**Describe the bug**
I want to be able to scale an item and keep one axis static.

**Expected behavior**
My suggestion is to do how many other editors do and allow the user special controls when they drag a resize handle while holding down the Ctrl key. For example, I could hold down Ctrl, grab the resize handle, and it would drag along the X axis or Y axis (whichever's closest).

SVG-Edit Version: 5.0.0

I may be able to fulfill this request - I haven't taken a look at your source yet, but I'd love to contribute to such a great project.

bstee615:
Let me know what you think, please. I think it would also be good to add a small icon that shows the "anchor point" for the axis-lock, the point that is assigned when the user presses ctrl. Otherwise, the scale of the object can jump around a lot for what seems like no reason.

Fixes #346.


editor/svgcanvas.js
```diff
@@ -2254,28 +2263,29 @@ const mouseMove = function (evt) {
       dy = r * Math.sin(theta);
     }
 
+    var normalized_dx = dx, normalized_dy = dy;
```
brettz9: Please use `const` where possible and `let` otherwise (and please run `npm run eslint` to make sure your PR doesn't add any new linting issues such as this).

index.html
```diff
@@ -0,0 +1,2 @@
+<iframe src="editor/svg-editor.html?extensions="
+    width="100%" height="100%"></iframe>
```
brettz9: I don't know that we want this for all projects. But feel free to add `index.html` to `.gitignore` as a kind of placeholder.

editor/svgcanvas.js
bstee615: Thank you for the feedback. While we are on the topic of code quality, was there an area where documentation was lacking? I will try to sprinkle some comments into my revisions where they're appropriate.

index.html
bstee615: No, I had just used this to run the application locally.

editor/svgcanvas.js
brettz9: You could add mention of your addition to `docs/tutorials/Editor.md`, and this file could in general be improved as well (in a separate PR), as there is precious little aimed at users uninitiated in graphics editors to svgedit.

In this besides a mention of the key-sequence and behavior in `Editor.md`, it'd be ideal if you could also add a ui test, as that is, AFAIC, the gold standard for testing the app, even while unit tests are still useful.

index.html
brettz9: Ok, so if you can remove this from the PR then.

editor/svgcanvas.js
bstee615: I don't think that improvements to `Editor.md` fit into the scope of this PR, but it'd be something to look at. Since that file is so slim right now, it would take a lot of groundwork to explain each tool that the ctrl-click applies to, so I'll leave this out of this PR.

index.html
bstee615: Ok!

bstee615:
Hi, I addressed the comments you left. If you have any other issues with the style, it would be good to point them out so I can pull them into this PR, but otherwise, I'd rather complete this one as I see the code as well documented. Thanks! Give it a look when you have a chance, please.

brettz9:
Sorry, still meaning to get back to you on this... Hopefully soon...

editor/svgcanvas.js
brettz9: General improvements would not fit into scope, but I would like to enforce a policy of documenting any new UI features, so I would appreciate it if you'd just describe the functionality you added/changed.

You don't have to leave the groundwork there for the tools (except maybe to add a "(To-do)" block in the sections you think need addition. Having these updated docs should both help encourage us to do this gradually, and be there to inform anyone familiar with older behavior (e.g., if the earlier developers were to resume work).

Likewise if you can add a Testcafe test--of if you need help with it, give me a human-readable series of steps here (with screenshots if necessary), and I can try to provide the Testcafe code. (Though I may switch us to Cypress as they know have UI testing code coverage, which is really awesome.) I am not even into design myself, so I should be a good sounding board to ensure the steps are clear.

Actually, the tests are even more important than docs, not only for regressions, but also as it is itself a form of documentation.

brettz9:
So if you could provide a UI test, or just a human-readable script, e.g.:

1. Hold "Ctrl" key
2. Click on coordinate 50, 100 and drag to 100, 200
3. Verify that image icon element XYZ exists...

bstee615:
Hi Brett, things got a lot busier since this time and to be honest, I forgot about this. Right now, I don't have time to re-set-up the env and add the test/manually test this PR. I'll close it for now and I hope you can use something from it. Thanks so much for your work. Cheers!

Issue #317, opened 2019-01-14
"Open SVG" issues


Nefiji:
If I click on "Open SVG" in the SVG-Edit drop-down menu, and then select a .svg file from my computer, SVG-Edit doesn't correctly open the file on its own. Only after I click on "Open SVG" again, it proceeds to open the .svg file from my first attempt. (Without having to select a file again) But even then, it pops up the "Loading image, please wait..." message indefinitely. Although when clicking on "Cancel", it clearly shows that the SVG already has been loaded completely and without an issue.

I don't know why this is happening, but it's happening on both of my browsers (Chrome & Firefox) - without any plugins interfering. Tell me if you need any further information.

brettz9:
IIRC, this may be an issue with not allowing pop-ups for the domain you are using... Can you try that and see if that helps?

brettz9:
Closing as no reply. Feel free to reopen if you have more info.

PR #351, opened 2019-08-03
Use change event to be sure that file is loaded before the function

Moliman:
Click event can trigger the script before the user choose the file, so it will be loader on the next
Open SVG click. Change event trigger the script after the user choose the file.

This fix the issue #317


brettz9:
Thanks for the PR!

Issue #364, opened 2019-11-21
IE: After draw Rectangle there is no select event happen


fkcailiniyazi:
**Describe the bug**
After I drawed a Rectangle, I tried to select it by click, but it didn't work, I have to click selection from toolbar or click resize/rotate to select it.

**To Reproduce**
1. Open SVGEditor from Internet Explorer
2. Draw a rectangle
3. Click on it
4. See the error

**Expected behavior**
Like other tool(circle for example) after click on it, a rectangle should be selected.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Internet Explorer 10

**Screenshots**

![Screenshot (3)](https://user-images.githubusercontent.com/54360982/69351885-474e1900-0c7c-11ea-8318-5205bb018f68.png)

brettz9:
Are you able to confirm your fix still works in other browsers besides IE?

fkcailiniyazi:
I have tested with latest versions of Chrome/Opera/Firefox/Microsoft Edge/ Internet Explorer 10, 11, this works.

brettz9:
Ok, great.

Would you mind also adding a UI test case (to the just-now-added to `master`  Cypress UI tests)? Specifically, if you could add to `cypress/integration/ui.js`.

(Cypress is a UI testing tool which has the awesome ability to give us coverage reports (and its coverage can also integrate with regular unit tests, though our old HTML tests have not yet been converted over). It sadly does not support non-Chromium based browsers at the moment, but I thought it was worth switching from Testcafe due to this great ability to get coverage info.)

fkcailiniyazi:
I am sorry that I'm late to reply to you, I was at vacation.

I could try to add UI test case, but I don't know how to do it, I checked the documentation, but I found difficulties to understand it.

brettz9:
I will need more time then, as I have a number of projects on my plate. If you look at the examples though (including the ones in our project), Cypress is pretty intuitive to pick up.

brettz9:
Even though Cypress [doesn't yet support other non-Chrome browsers](https://github.com/cypress-io/cypress/issues/310), I'd still like to add the test to ensure we're ok with other browsers, and once they may get that support.

In trying to prepare the UI case, however, I'm not clear on what happens after you draw the rectangle. Are you saying it is not selected right after drawing the rectangle?

Also, when you say "click resize/rotate", which form controls are you talking about? Don't the resize/rotate controls only show up after the rectangle is selected?

[...]

PR #366, opened 2019-11-21
fixed bug of issue#364

fkcailiniyazi:
fixed bug of issue #364

### DIFF
--- a/dist/index-es.js
+++ b/dist/index-es.js
@@ -15363,7 +15363,9 @@ function SvgCanvas(container, config) {
         startY = snapToGrid(startY);
       } // if it is a selector grip, then it must be a single element selected,
       // set the mouseTarget to that and update the mode to rotate/resize
-
+      if (mouseTarget.tagName == 'rect' && !isNullish(selectedElements[0]) && currentMode !== 'select') {
+        currentMode = 'select';
+      }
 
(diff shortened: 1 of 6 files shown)


cuixiping:
You need not commit the `dist/*.*` and `*-iife.js` files.

brettz9:
Yeah I guess since our Github Pages version is a separate repo, we need only have that repo include them (and I have just now updated that repo's .gitignore to avoid copying over some testing/coverage code which is I think not necessary to have online).

Feel free to add a commit to our regular svgedit repo to add `dist`, etc. to `gitignore` (though not to `.npmignore`), as the https://github.com/SVG-Edit/SVG-Edit.github.io fork will not copy each release's `.gitignore`; so as long as the `dist` files are built, they will be copied over and can be added to the repo when running the fork's copy script.

brettz9:
And FWIW, at one point there were Git hosting libraries, so we could demo the latest on `master`, making inclusion of the `dist` files useful, but at this point, they are now apparently all offline.

cuixiping:
Latest master demo should use source file only, not the built files on dist directory.

brettz9:
It is nice to have some sanity checking to make sure the dist files are working, but I think such a change would be ok.

jfhenon:

Issue #381, opened 2020-01-30
Feature: specify viewport units other than pixels (inches, meters, etc...)


klesun:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%242%20Funded-%2300A156.svg" />](https://issuehunt.io/r/SVG-Edit/svgedit/issues/381)

**Describe the bug**
It's actually a feature request. Posting since nobody requested it yet.
When you upload an svg having viewport width and height in millimeters: `width="60mm" height="90mm"`, svg-edit automatically converts them to pixels. Whereas inkscape for example preserves the units from the input.

I do get that supporting other units than pixels is potentially a huge amount of work, but it would be nice to have the support in `svg-edit` since units are part of [svg standard](https://www.w3.org/TR/SVG/coords.html).

**To Reproduce**
1. Upload following svg with `svgEditor.loadFromString()` (or open it with _SVG-Edit -> Open SVG_):
    ```xml
    <?xml version="1.0"?>
    <svg width="60mm" height="90mm" xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
     <!-- Created with SVG-edit - https://github.com/SVG-Edit/svgedit-->
     <g class="layer">
      <title>Layer 1</title>
      <image height="400" id="svg_1" width="614" x="13" xlink:href="https://picsum.photos/id/433/614/400" y="35"/>
     </g>
    </svg>
    ```
2. Go to _SVG-Edit -> Document Properties_
3. See that input `width="60mm" height="90mm"` was converted to `226.7716535433071` and `340.15748031496065`

**Expected behavior**
I'd like the width and height to stay in original units - `mm` and have a dropdown like in inkscape to choose the units to convert to. Would be nice to have the rulers also in the selected units.

**SVG-Edit environment (IMPORTANT)**
- File for SVG-Edit: `svg-editor.html`
- Version: 5.1.0
- Protocol: both http (on my local server) and https (in your demo)

**Desktop (please complete the following information):**
 - OS: Linux Mint 18.3
 - Browser Chromium 
 - Version v79.0.3945.79

**Screenshots**
Document properties in `svg-edit`:
![image](https://user-images.githubusercontent.com/5202330/73436185-b61db180-4352-11ea-93d9-63e18fad3b10.png)

Document properties in `inkscape`:
[...]

PR #384, opened 2020-02-08
Enhancement: #381 Set base units preference from input svg file header

klesun:
I took few pretty daring decisions in order to achieve the desired result, any improvement suggestions are very welcome.

The thing that will most likely arise your concern is that I changed return value of `svgCanvas.setSvgString()` from `true/false` to `Object/false`, where object holds the parsed SVG values that should be passed to the editor.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ mochawesome-report
 mochawesome.json
 
 .vscode
+/.idea/
--- a/editor/svg-editor.js
+++ b/editor/svg-editor.js
@@ -333,9 +333,10 @@ let svgCanvas, urldata = {},
  *   loading fails and `noAlert` is truthy.
  */
 async function loadSvgString (str, {noAlert} = {}) {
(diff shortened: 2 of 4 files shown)


brettz9:
In order to expand our test coverage, and to prevent regressions, I've started asking for (Cypress) UI tests to be added for new changes. Would you be able to add a test for this (within `cypress/integrations`)?

Btw, for those who may have been concerned about Cypress having been limited to Chrome, Cypress has now added Firefox beta and Edge support along with its Chrome support.

klesun:
Ok, thanks for response, will add a Cypress test. Will take some time, though (never worked with Cypress before).

klesun:
In the end we did not need these changes in our project, and after I thought about it for some time, for a regular use case these changes are pretty arguable. And I sadly have no time at all to follow up, so I guess I'll beter close this useless PR. Sorry for the noise.

PR #389, opened 2020-02-21
Move clipboard from localStorage to sessionStorage

### LINKED ISSUE #379: Clipboard uses local storage and persists forever
**Describe the bug**
SvgCanvas uses `localStorage` to save the contents of the clipboard.  This lasts forever, and has GDPR implications since users do not expect that the contents of their clipboard will survive beyond the current session, let alone a reboot of their computer.

**To Reproduce**
1. Go to https://svg-edit.github.io/svgedit/releases/latest/editor/svg-editor.html
2. Create any shape.
3. Cut the shape using the keyboard shortcut or the context menu.
4. Close the tab, close your browser (optionally reboot your computer).
5. Restart your browser, reopen a tab to https://svg-edit.github.io/svgedit/releases/latest/editor/svg-editor.html
6. Paste using the keyboard shortcut or the context menu.

**Expected behaviour**
The expected behaviour is that nothing would be pasted.  The actual behaviour is that the content is pasted.

**Proposed solutions**
I'm happy to implement the fix, but I want opinions on which direction to pursue before doing so.
* Switching from localStorage to sessionStorage would be trivial.  However, it would not allow for copy/paste operations between tabs.
* Switching from localStorage to a session cookie would be nearly as trivial and does work across tabs.  However, cookies would impose a 4 kb limit on the clipboard, instead of the storage API's 5 MB limit.
* Switch to sessionStorage, but also add event monitoring to synchronize data across tabs.  See [this answer on StackOverflow](https://stackoverflow.com/questions/20325763/browser-sessionstorage-share-between-tabs).

NeilFraser:
This change eliminates the permanent clipboard that unexpectedly survives beyond the current session.  Uses localStorage to momentarily flash the clipboard to other tabs.  Resolves #379
Also fixes traceback if paste is called with nothing on clipboard.
Also fixes traceback if copy is called and no context menu exists.
Adds Cypress test for clipboard.


brettz9:
Great to see this, thanks! Look forward to reviewing...

Btw, in future, feel free to keep your old PRs open and just amend or add to them, including new comments.

NeilFraser:
Yes, the branch the old PR was on suffered catastrophic failure, and GitHub auto-closed the PR without my input.

brettz9:
I think the code and testing looks good. (I've also pushed to this branch a few unrelated items improving our test linting.)

However, on further reflection, I do think we ought to keep the option to use `localStorage`. I like the fact that the work is still there the next time I load the page. It is kind of like a built-in auto-save. Do you think you could swap the session storage based on a preference setting?

Even better might be having separate extensions, one for session storage (a default-enabled extension) and one for local storage (and no extension just tracking in memory).

What do you think?

brettz9:
Sorry--lack of sleep here. I was thinking of saving rather than clipboard.

brettz9:
Still, if it survives session crashes, that is a good thing to me. What do you think about still making it a preference?

NeilFraser:
> Still, if it survives session crashes, that is a good thing to me.

SessionStorage does survive crashes.  Chrome and Firefox will restore tabs with all session data intact (tested just now).

> How is the undo history stored at the moment?

Undo history is stored as a local array in history.js: `this.undoStack = [];`  There is no persistence.

LocalStorage is pretty scary given that it *never* expires.  In today's world any unexpected incident of data retention (even accidental) can result in a crippling fine.  €20 million for individuals or small companies, or in the case of my employer, €6 billion.

brettz9:
Good to know re: crashes, thanks very much. And it looks like `sessionStorage` may survive when "Recently closed tabs" are restored, so in the many cases where I accidentally close tabs prematurely, I can get back at them.

Even more than the clipboard, it would be especially nice as a separate feature to preserve the Undo Stack across sessions. That is something I very much appreciate being restored (and surviving beyond sessions). But not needed here.

I understand about data retention concerns, and support configuring to allow restricting this, but the default behavior now in SVG-edit is to explicitly ask for user permission to store data locally (though not sure those settings were being checked in the case of the clipboard).

Anyways, retaining the clipboard, while nice to have the option, is not critical, and indeed it is better as a default to retain it in the session, so I went ahead and merged. For future storage changes though, let's allow setting and checking of a user preference; with the API shared between the interfaces, it shouldn't be so hard to implement.

Btw, I'm trying to get a release ready, but having trouble with getting the Git submodule syntax right. (Whenever I try updating, it either grabs from `master` or complains about the branch missing (though it is present on SVG-Edit/svgedit) rather than the branch pointed to by the submodule in `.gitmodules`.)

Issue #407, opened 2020-04-19
Double click is not working to edit text that is a direct child of a group


jfhenon:
**Describe the bug**
Double click is not working to edit text within a group.

**To Reproduce**

1.Load the following SVG: 
````
<svg width="640" height="480" xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg">
 <!-- Created with SVG-edit - https://github.com/SVG-Edit/svgedit-->
 <g class="layer">
  <title>Layer 1</title>
  <g id="svg_3">
    <rect fill="#ffff00" height="70" id="svg_1" stroke-dasharray="null" stroke-linecap="null" stroke-linejoin="null" stroke-width="0" width="165" x="179.5" y="146.5"/>
    <text fill="#000000" font-family="Arial" font-size="40" id="svg_2" stroke-dasharray="null" stroke-linecap="null" stroke-linejoin="null" stroke-width="0" text-anchor="middle" x="260.5" xml:space="preserve" y="192.5">hello</text>
   </g>
 </g>
</svg>
````
2. No way to move into textedit mode to change  "hello" into "bonjour" for example when you double click on the word "hello"
**Expected behavior**
Dble click should put you in text edit
**SVG-Edit environment (IMPORTANT)**
https://unpkg.com/svgedit@5.1.0/editor/svg-editor.html

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Chrome
 - Version  81.0.4044.113

**Screenshots**
 
![image](https://user-images.githubusercontent.com/20402845/79684246-11fc2600-8230-11ea-920d-5c24679262b8.png)

**Additional context**

I think the error is related to this line:
https://github.com/SVG-Edit/svgedit/blob/ced0d2d778effab3b428dd8d3d2104f9439e9411/editor/svgcanvas.js#L2893

Indeed, the double click will return immediately so the line:
https://github.com/SVG-Edit/svgedit/blob/ced0d2d778effab3b428dd8d3d2104f9439e9411/editor/svgcanvas.js#L2898
will not have a chance to put the user in the edit mode which is the intention if you double click on the word. 

I could write directly a PR but I'm not yet familiar with the whole code so maybe next time...

Issue #408, opened 2020-04-19
Bug when trying to show SVG content


jfhenon:
**Describe the bug**
Bug when trying to show SVG content when the context is changed.

**To Reproduce**
1. Load the following sampe SVG:
```
<svg width="640" height="480" xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg">
 <!-- Created with SVG-edit - https://github.com/SVG-Edit/svgedit-->
 <g class="layer">
  <title>Layer 1</title>
  <g id="svg_6">
   <rect fill="#FF0000" height="71" id="svg_1" stroke="#000000" stroke-width="5" width="94" x="69.5" y="51.5"/>
   <circle cx="117.5" cy="87.5" fill="#ffff00" id="svg_5" r="19.84943" stroke="#000000" stroke-dasharray="null" stroke-linecap="null" stroke-linejoin="null" stroke-width="5"/>
  </g>
 </g>
</svg>
````
2. move the context to the group by double-click on the rectangle
![image](https://user-images.githubusercontent.com/20402845/79689039-93fd4680-8252-11ea-9218-a7372ce80d79.png)
3. click on the following button ![image](https://user-images.githubusercontent.com/20402845/79689115-08d08080-8253-11ea-8455-fa9c3a0dab72.png) 
4. the action will fail and an error is displayed in the console: 
![image](https://user-images.githubusercontent.com/20402845/79689102-ed657580-8252-11ea-8430-c7593e352856.png)
**Expected behavior**
A window with the SVG content should be displayed.

**SVG-Edit environment (IMPORTANT)**
https://unpkg.com/svgedit@5.1.0/editor/svg-editor.html

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Chrome
 - Version 81.0.4044.113

**Screenshots**
see above

**Log**
see above 

**Additional context**
https://github.com/SVG-Edit/svgedit/blob/4b92acc79047ad3fede9945c3bc0a55f8c23f6ef/editor/svgcanvas.js#L695

This seems to related to the conditional statement above. If showGrip is false or undefined AND selectedElements.length == 0, it fails on the statement selectorManager.requestSelector. 

should it be the following? 
```
if (selectedElements.length === 1) {
  selectorManager.requestSelector(selectedElements[0]).showGrips(showGrips)
}
```

brettz9:
Hi @jfhenon .... Is there anything left outstanding on this?

PR #416, opened 2020-06-19
Fix: issue #407 and #408

jfhenon:


brettz9:
Hi... Great to have your contributions. But could you add a Cypress test or tests to both document the behavior we are now expecting and to prevent future regressions. Thanks!

jfhenon:
> Hi... Great to have your contributions. But could you add a Cypress test or tests to both document the behavior we are now expecting and to prevent future regressions. Thanks!

Added!

brettz9:
This looks great, thanks! I hope to review within the next couple of days, though feel free to ping me if not. If we can get test coverage, I really think this should bring added clarity to intended behaviors, prevent regressions, and be inviting to would-be contributors.

Issue #417, opened 2020-06-19
Shortcuts with imagelib


EetuHakkinen:
**Describe the bug**
Keyboard shortcuts does not work after importing image from image library.

**To Reproduce**

1. Go to svgedit menu -> image library
2. Select image library
3. Select image
4. Now keyboard shortcuts is not working
5. When clicking outside drawing area (for example menu bar at the top or layers menu at the right), keyboard shortcuts starts to work again.

SVG-Edit environment (IMPORTANT)

File for SVG-Edit: https://unpkg.com/svgedit@5.1.0/editor/svg-editor.html
Desktop (please complete the following information):

OS: windows10
Browser: Chrome Version 83.0.4103.106 (Official Build) (64-bit)

brettz9:
This seems to be Chrome-specific; I can't replicate on Firefox. The project is not being actively developed, though PRs with tests are welcome.

PR #421, opened 2020-07-04
Fix issue #417

jfhenon:
## PR description

fix issue #417 and took advantage of this to move shortcut code to native ES6/dom code.

## Checklist

- [!] - Added Cypress UI tests **-> UI test was not added as iFrame is not supported yet by Cypress and it is needed to test the issue.**
- [ n/a] - Added any user documentation. Though not required, this can be a big

### DIFF
--- a/editor/svg-editor.js
+++ b/editor/svg-editor.js
@@ -5626,13 +5626,20 @@ editor.init = function () {
     const toolButtons = [
       {sel: '#tool_select', fn: clickSelect, evt: 'click', key: ['V', true]},
       {sel: '#tool_fhpath', fn: clickFHPath, evt: 'click', key: ['Q', true]},
-      {sel: '#tool_line', fn: clickLine, evt: 'click', key: ['L', true], parent: '#tools_line', prepend: true},
-      {sel: '#tool_rect', fn: clickRect, evt: 'mouseup', key: ['R', true], parent: '#tools_rect', icon: 'rect'},
-      {sel: '#tool_square', fn: clickSquare, evt: 'mouseup', parent: '#tools_rect', icon: 'square'},
-      {sel: '#tool_fhrect', fn: clickFHRect, evt: 'mouseup', parent: '#tools_rect', icon: 'fh_rect'},
-      {sel: '#tool_ellipse', fn: clickEllipse, evt: 'mouseup', key: ['E', true], parent: '#tools_ellipse', icon: 'ellipse'},
-      {sel: '#tool_circle', fn: clickCircle, evt: 'mouseup', parent: '#tools_ellipse', icon: 'circle'},
-      {sel: '#tool_fhellipse', fn: clickFHEllipse, evt: 'mouseup', parent: '#tools_ellipse', icon: 'fh_ellipse'},
+      {sel: '#tool_line', fn: clickLine, evt: 'click', key: ['L', true],
+        parent: '#tools_line', prepend: true},
+      {sel: '#tool_rect', fn: clickRect, evt: 'mouseup',
+        key: ['R', true], parent: '#tools_rect', icon: 'rect'},
+      {sel: '#tool_square', fn: clickSquare, evt: 'mouseup',
+        parent: '#tools_rect', icon: 'square'},
+      {sel: '#tool_fhrect', fn: clickFHRect, evt: 'mouseup',
+        parent: '#tools_rect', icon: 'fh_rect'},
+      {sel: '#tool_ellipse', fn: clickEllipse, evt: 'mouseup',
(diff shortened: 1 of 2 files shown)


editor/svg-editor.js
```diff
@@ -5626,20 +5626,13 @@ editor.init = function () {
...
-      {sel: '#tool_fhellipse', fn: clickFHEllipse, evt: 'mouseup',
-        parent: '#tools_ellipse', icon: 'fh_ellipse'},
+      {sel: '#tool_line', fn: clickLine, evt: 'click', key: ['L', true], parent: '#tools_line', prepend: true},
+      {sel: '#tool_rect', fn: clickRect, evt: 'mouseup', key: ['R', true], parent: '#tools_rect', icon: 'rect'},
+      {sel: '#tool_square', fn: clickSquare, evt: 'mouseup', parent: '#tools_rect', icon: 'square'},
+      {sel: '#tool_fhrect', fn: clickFHRect, evt: 'mouseup', parent: '#tools_rect', icon: 'fh_rect'},
+      {sel: '#tool_ellipse', fn: clickEllipse, evt: 'mouseup', key: ['E', true], parent: '#tools_ellipse', icon: 'ellipse'},
+      {sel: '#tool_circle', fn: clickCircle, evt: 'mouseup', parent: '#tools_ellipse', icon: 'circle'},
+      {sel: '#tool_fhellipse', fn: clickFHEllipse, evt: 'mouseup', parent: '#tools_ellipse', icon: 'fh_ellipse'},
       {sel: '#tool_path', fn: clickPath, evt: 'click', key: ['P', true]},
```
brettz9, line 5636: Though I've disabled it for now, I do have the `max-len` rule in `.eslintrc.js` to try to enforce this. For smaller screen widths, and for the sake of those not wanting to always use wrapping in their IDEs (since it can be less clear when something is wrapping), I'd personally prefer keeping the list as it was. While it can be nice on larger screens to take advantage of getting more in view, I think it is more cumbersome to have to scroll horizontally.

editor/svg-editor.js
```diff
@@ -5796,42 +5786,37 @@ editor.init = function () {
...
         });
+        // register the keydown event
+        document.addEventListener('keydown', (e) => {
+          const key = `${(e.metaKey) ? 'meta+' : ''}${(e.ctrlKey) ? 'ctrl+' : ''}${e.key.toLowerCase()}`;
+          if (!keyHandler[key]) return;
+          keyHandler[key].fn();
+          if (keyHandler[key].pd) {
+            e.preventDefault();
+          }
+        });
```
brettz9, line 5819: Great to see us moving away from the dependencies!

editor/svg-editor.js
```diff
@@ -5759,20 +5748,21 @@ editor.init = function () {
...
+        const keyHandler = {}; // will contain the action for each pressed key
 
-        $.each(toolButtons, function (i, opts) {
+        toolButtons.forEach((opts) => {
           // Bind function to button
           let btn;
           if (opts.sel) {
-            btn = $(opts.sel);
-            if (!btn.length) { return true; } // Skip if markup does not exist
+            btn = document.querySelector(opts.sel);
```
brettz9, line 5757: I think this really needs to be in a utility. If you import my Jamilih library (already a part of the project elsewhere), it exports `$` and `$$` for querySelector and querySelectorAll (which you could rename to avoid jQuery conflicts).

jfhenon: I need to go deeper into your library but, for now,  I like the idea that every developper would quickly understand what the line does so  I'd rather keep native functions at least until we get rid of jQuery.

brettz9: In Jamilih, it is implemented with just these lines:

```js
const $ = (sel) => doc.querySelector(sel);
const $$ = (sel) => [...doc.querySelectorAll(sel)];
```

However, you are most welcome to add copies of these locally into our utilities file so they are more immediately traceable as to their implementation. But I think it is too frequent of a need not to have it as a utility in *some* form, and having `querySelector` explicit everywhere, while semantic, is just too verbose and distracting imo for the ever-present need for finding items by selector. Developers can follow the import for this to find its implementation as with other functions in the library. For pretty much any other function I agree it is better to directly use native functions when not significantly more verbose.

jfhenon: OK, I agree to do this but when we will have no more jQuery in the project so $ can be used without the risk to mix with jQuery $

brettz9: Ok... how does `$q` and `$qq` sound?

jfhenon: Ok for me.

PR #427, opened 2020-08-06
New build

jfhenon:
## PR description
This PR has the goal to simplify the build process using the latest rollup/babel updates offering modern browsers light solution but also keeping support for older browsers.
We may also include in this PR some npm packages that are now imbedded in the project and not updated (which ultimately can cause bugs and security issues).

**NOT READY TO MERGE BUT FOR INITIAL COMMENTS/CONTRIBUTIONS**

## Checklist


jfhenon:
Note that we should land in the documentation: 
- svgedit.html is now index.html (es version) so you can just type: http://localhost:8000/dist/editor
- other versions can be accessed with a different html: http://localhost:8000/dist/editor/index-iife.html or http://localhost:8000/dist/editor/index-umd.html
- The source file can also be tested using http://localhost:8000/src/editor
- build is just `npm run build `(the rollup config will do the rest)

jfhenon:
Now, css and modules are added using standard css3/es6 syntax. It works well with esm but not tested on old browser.

jfhenon:
> I haven't dived too deeply into the details of the Rollup process yet, as I'd like to request these more ancillary changes first and then if all the editors are ready to be tested, I can check at that point.
> 
> But it generally looks good and more elegant if it meets the needs of all of our editors, and if the issues I raised can be addressed.

Yep, this was my intention to have a brief feedback from you. This will also takes some time before I can complete this.

package.json
```diff
@@ -101,9 +93,9 @@
     ]
   },
   "lint-staged": {
-    ".eslintignore": "npm run eslint-arg --",
-    ".eslintrc.js": "npm run eslint-arg --",
-    "src/**/*.{js,md,html}": "npm run eslint-arg --",
+    ".eslintignore": "npm run eslint",
+    ".eslintrc.js": "npm run eslint",
+    "src/**/*.{js,md,html}": "npm run eslint",
```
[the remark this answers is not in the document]
jfhenon: Yes, I disabled it as it was causing an issue on my environment but I agree lint-staged is a nice addition.

src/editor/spinbtn/jQuery.SpinButton.js
```diff
@@ -75,13 +75,6 @@
  * @returns {external:jQuery}
 */
 export default function jQueryPluginSpinButton ($) {
-  if (!$.loadingStylesheets) {
-    $.loadingStylesheets = [];
-  }
-  const stylesheet = 'spinbtn/jQuery.SpinButton.css';
-  if (!$.loadingStylesheets.includes(stylesheet)) {
-    $.loadingStylesheets.push(stylesheet);
-  }
```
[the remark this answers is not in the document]
jfhenon: The current code is not really "unaware" that spinButton is not used, look svgedit.js: 
````
function getStylesheetPriority (stylesheetFile) {
          switch (stylesheetFile) {
          case 'jgraduate/css/jPicker.css':
            return 1;
          case 'jgraduate/css/jGraduate.css':
            return 2;
          case 'svg-editor.css':
            return 3;
          case 'spinbtn/jQuery.SpinButton.css':
            return 4;
          default:
            return Infinity;
          }
````

More generally, I'm convinced it's more maintainable, modular and extensible to use **standard** ways to include css  stylesheets...
Actually, the more common/recent way to do this is to directly include the css in the js file importing SpinButton so you actually only include the CSS if spinButton is actually used. You can then use rollup packages such as rollup-plugin-css-only. 

I didn't choose this approach because it would prevent to use the src directly with Chrome (as it is not supported to import a css directly yet). We may adopt this approach later because it's possible to use rollup/babel in watch mode and to use the dist folder for tests and distribution for all editor versions.

File: src/editor/svg-editor.manifest [hunk 3]
```diff
@@ -1,156 +0,0 @@
...
-extensions/ext-panning.js
-extensions/ext-php_savefile.js
-extensions/ext-polygon.js
-extensions/ext-server_moinsave.js
-extensions/ext-server_opensave.js
-extensions/ext-shapes.js
-extensions/ext-star.js
-extensions/ext-storage.js
-extensions/ext-webappfind.js
-extensions/ext-xdomain-messaging.js
```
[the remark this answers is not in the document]
jfhenon: Ok

src/editor/svgedit.js
```diff
@@ -371,11 +350,11 @@ function getImportLocale ({defaultLang, defaultName}) {
...
      * @returns {Promise<module:locale.LocaleStrings>} Resolves to {@link module:locale.LocaleStrings}
      */
-    function importLocale (language) {
+    async function importLocale (language) {
       const url = `${curConfig.extPath}ext-locale/${name}/${language}.js`;
-      return importSetGlobalDefault(url, {
-        global: `svgEditorExtensionLocale_${name}_${language.replace(/-/g, '_')}`
-      });
+      // eslint-disable-next-line node/no-unsupported-features/es-syntax
+      const locale = await import(url);
```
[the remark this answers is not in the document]
jfhenon: It's incomplete now but this should work. By the way, how do you test old browsers?

Other more generic question: many applications tends to drop support to old browsers and to suggest to use an older version of svgedit if they have no choice (svgedit 5.x for us)?

I think we are adding a lot of complexity to maintain svgedit because if this support for old browsers (which by the way cause security concerns).

src/xdomain-svgedit-config-es.js
```diff
@@ -1,6 +0,0 @@
-import svgEditor from './editor/svg-editor.js';
-
-svgEditor.setConfig({
-  canvasName: 'xdomain', // Namespace this
-  allowedOrigins: ['*']
-});
```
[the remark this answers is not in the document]
jfhenon: Yes, not yet done but I agree with your point.

src/svgcanvas/svgcanvas.js
```diff
@@ -4076,24 +4074,12 @@ this.exportPDF = async function (
...
-        ? 'module'
-        : 'text/javascript'
-    });
-    // await importModule('jspdf/jspdf.plugin.svgToPdf.js');
+    // eslint-disable-next-line node/no-unsupported-features/es-syntax
+    await import('../editor/jspdf/underscore-min.js');
+    // eslint-disable-next-line node/no-unsupported-features/es-syntax
+    await import('../editor/jspdf/jspdf.min.js');
+    // eslint-disable-next-line node/no-unsupported-features/es-syntax
+    await import('../editor/jspdf/jspdf.plugin.svgToPdf.js');
```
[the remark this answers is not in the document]
jfhenon: Agree

package.json
```diff
@@ -161,28 +153,38 @@
     "not Safari < 10",
     "not ios_saf < 10"
   ],
-  "dependencies": {},
+  "dependencies": {
+    "@babel/polyfill": "^7.10.4",
+    "core-js": "^3.6.5",
+    "regenerator-runtime": "^0.13.7"
+  },
```
brettz9, line 160: The reason I had `core-js-bundle and `regenerator-runtime` in devDependencies and bundled these through a copy routine, is so that our demo would work on GitHub Pages (since it doesn't install npm dependencies).

There should be no need for `@babel/polyfill`, as that is deprecated by Babel and its replacements are already present (core-js and regenerator-runtime).

package.json
```diff
@@ -223,14 +230,12 @@
     "promise-fs": "^2.1.1",
     "qr-manipulation": "git+https://github.com/brettz9/qr-manipulation.git",
     "query-result": "git+https://github.com/WebReflection/query-result.git",
-    "react": "^16.13.1",
-    "react-dom": "^16.13.1",
```
brettz9, line 232: While I know it can sometimes be faster to just change things and submit , I tend to have reasons for including things, so you might want to email me to ask why something that seems unecessary is in place before removing in case I don't catch it in my review, and then we have more back-and-forth after we later rediscover the need.

While we strictly speaking don't need these, we get a bunch of warnings with them on installation (which may distract from other more serious warnings that could occur) since some of our linting plugins have them as peer dependencies (I've sought changes in the responsible plugin so it won't be necessary, but still awaiting a reply).

dist/editor/index.html
```diff
@@ -23,19 +23,9 @@
     <script src="jquery.min.js"></script>
     <!--{endif}-->
 
-    <!-- Lacking browser support -->
-    <script nomodule="" src="redirect-on-no-module-support.js"></script>
-    <script type="module" src="redirect-on-lacking-support.js"></script>
-
```
brettz9, line 29: I had these as separate scripts so that:

1. The user whose browser did not support would quickly be notified without need to load the whole app.
2. It wouldn't pollute the main code. While I normally believe in modularity, for old-browser polyfills and such detection, I think it does make more sense to have separate scripts since one can review browser support in one place--i.e., if one is updating browser targets, one can just look in the HTML and drop all polyfills that are no longer needed without a need to dig into the JavaScript files and modify them based on a concern unrelated to a change in behavior in those scripts (though it is still good to add comments indicating the files where the polyfills are needed). Admittedly this could be mitigated if there is an app entrance file separate from the main editor file which handles the importing of polyfills all in one place, but there'd still be problem no. 1.

dist/editor/index.html
```diff
@@ -23,19 +23,9 @@
...
     <!-- As yet no ES6 Module support -->
     <script src="jquery-ui/jquery-ui-1.8.17.custom.min.js"></script>
 
-    <!-- Browser polyfills -->
-    <script src="../external/dom-polyfill/dom-polyfill.js"></script>
-
-    <!-- ES6+ polyfills (Babel) -->
-    <script src="../external/core-js-bundle/minified.js"></script>
-    <script src="../external/regenerator-runtime/runtime.js"></script>
```
brettz9, line 29: See comment just above. (These don't suffer from problem no. 1 though, so the mitigation mentioned at the end of no. 2 could work for these polyfills if you wanted to switch them to imports.)

src/editor/index.js
```diff
@@ -18,6 +18,10 @@ For default config and extensions (and available options) available to
 `setConfig()`, see the file `docs/tutorials/ConfigOptions.md`
 */
 
+// eslint-disable-next-line node/file-extension-in-import
+import 'core-js/stable';
+// eslint-disable-next-line node/file-extension-in-import
+import 'regenerator-runtime/runtime';
```
brettz9, line 24: This style of import suffers from causing one not to be able to have working browser source since browsers only recognize relative paths (import maps can help with this, but it would be cumbersome currently: https://github.com/WICG/import-maps/issues/126 ). One has to go through the build step (or have a test server which processes these).

You could use `./node_modules` paths however, and the potential issue of non-flattened node_modules should be less likely given that svgedit is less likely to be a transitive dependency.

package.json
```diff
@@ -161,28 +153,38 @@
...
     "not ios_saf < 10"
   ],
-  "dependencies": {},
+  "dependencies": {
+    "@babel/polyfill": "^7.10.4",
+    "core-js": "^3.6.5",
+    "regenerator-runtime": "^0.13.7"
+  },
   "devDependencies": {
+    "@babel/cli": "^7.10.5",
```
brettz9, line 162: Are you sure this one is necessary? I wouldn't think so if doing through `@rollup/plugin-babel`.

jfhenon:
> The change to `import` sounds good if it will work (I can't remember why I didn't go this route). See my other comments.

I'm still struggling to find a way to configure Rollup properly for old browsers (nomodule). I think I progress slowly. I'm more familiar with webpack so would you have an issue if I test it on this one?

package.json
jfhenon: No. I don't think we need it. I was using it for tests. Please don't forget that this branch is far from being good to merge.

package.json
jfhenon: I think it is simpler and cleaner to have the demo only in dist (and we can copy or bundle dependencies during the build).

package.json
jfhenon: This branch is really in draft mode only. At this point, I'm not even sure that this is the correct way to refresh the build process. This is to explain that I will take every single of your comments before I consider the merge.

brettz9:
Ugh--I'm sorry but Webpack really has struck me the times I've had to use it as unnecessarily complex, and that seems to be a common view. Webpack also doesn't support ESM distributions--only ESM source.

What `nomodule` issue are you looking to address? One useful plugin is https://github.com/rollup/plugins/tree/master/packages/replace (or some other third party Rollup plugins that do replacing) which you might use to strip out some input in one build or another, though that is for within JS only I think (not sure if any of the HTML/asset plugins do replacements).

src/editor/index.js
jfhenon: See comment above. This branch is not yet GTM.

package.json
brettz9: I don't know for sure that GitHub Pages can allow any build process of its own.

jfhenon:
Ok. Let's stick to Rollup.

package.json
jfhenon: To my knowledge, the build needs to be done outside GitHub.

brettz9: Ok, I guess the important thing is just that the `dist`, including copies or bundled deps, is not gitignored. Then, yeah, that is simpler and cleaner.

brettz9: However, since it has to be bundled in `dist`, that brings us back to the fact that these should be moved back to `devDependencies`. Then regular users won't have to install these packages (they'll just get the `dist` copies installed--not all the `node_modules`).

.travis.yml
```diff
@@ -22,4 +22,4 @@ jobs:
         # - 12
         # - 10
       script:
-        - npm run eslint
+        - npm run lint
```
jfhenon, line 25: More standard way to call the linter

.babelrc.json
```diff
@@ -1,5 +0,0 @@
-{
```
jfhenon, line 1: replaced with the .js version

cypress.json
```diff
@@ -4,5 +4,7 @@
   "video": false,
   "reporterOptions": {
     "configFile": "mocha-multi-reporters.json"
-  }
+  },
+  "defaultCommandTimeout":	10000,
```
jfhenon, line 8: es-dev-server is transforming cjs to es at startup which causes a delay => longer timeout

jfhenon:
will reopen when branch is complete.

src/editor/index.js
jfhenon: changed the web server so src is usable (including for tests)

dist/editor/index.html
jfhenon: still need to check thos

src/editor/index.js
brettz9: Ok, but note that Netlify appears per https://community.netlify.com/t/support-guide-can-i-run-a-web-server-http-listener-and-or-database-at-netlify/3078 to be only for static files. So our demos wouldn't be able to work with the ESM editor (unless the editor used the ESM distribution files, and with source maps back to source so people reporting bugs could at least help us identify where the problems lay).

jfhenon: > Ok, but note that Netlify appears per https://community.netlify.com/t/support-guide-can-i-run-a-web-server-http-listener-and-or-database-at-netlify/3078 to be only for static files. So our demos wouldn't be able to work with the ESM editor (unless the editor used the ESM distribution files, and with source maps back to source so people reporting bugs could at least help us identify where the problems lay).

The demo hosted by Netlify will work from any of the bundle in the dist folder.

package.json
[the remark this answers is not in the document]
jfhenon: was fixed by Brett

src/editor/extensions/ext-mathjax.js
```diff
@@ -7,8 +7,6 @@
  * @copyright 2013 Jo Segaert
```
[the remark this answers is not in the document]
jfhenon: done

src/editor/svg-editor.manifest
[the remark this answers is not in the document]
jfhenon: move to archive for now.

src/xdomain-svgedit-config-es.js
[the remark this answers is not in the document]
jfhenon: done (using a rollup option)

package.json
jfhenon: removed

Issue #429, opened 2020-08-20
Layer window bug after importing an image


Quartzbell:
the layer window will be messed up when we import images through SVG Edit -> Import Image, or via drag&drop while the layer window is opened. The problem will persist until we reload the application.

![svg-edit-bug](https://user-images.githubusercontent.com/6446747/90720249-30168780-e2f1-11ea-88c2-571748fb4851.png)

PR #430, opened 2020-08-20
don't remove the style

jfhenon:
## PR description

Solve issue #429 

## Checklist


jfhenon:
@brettz9 the fix is working but I'm still unsure what was the initial intention of the line that I had to remove in the importImage function:  `$('#workarea').removeAttr('style');`

brettz9:
For bug fixes, just go ahead and address without reviews, but you will need to update the builds so that demos can work on `master` (I've been looking into getting our online demos restored). Of course, we want tests, but as long as you are on board to really start adding more of these, I don't mind some quick fixes so you are not too distracted from your main work.

PR #439, opened 2020-10-04
User extensions

jfhenon:
## PR description

in addition to standard extensions, this PR makes possible to add user extensions.they behave exactly like all extensions but they can be anywhere so the full relative path needs to specified for each user extensions. 
```
  userExtensions: [/* '../ext-helloworld/ext-helloworld.js' */]
```

## Checklist


.eslintrc.js
```diff
@@ -123,7 +123,8 @@ module.exports = {
     'import/no-anonymous-default-export': 'off',
     'node/no-unsupported-features/node-builtins': 'warn',
     'prefer-exponentiation-operator': 'warn',
-    'node/no-unsupported-features/es-syntax': 'off'
+    'node/no-unsupported-features/es-syntax': 'off',
+    'no-unsanitized/method': 'off'
```
brettz9, line 127: What was the reason for disabling this? I pretty much think there should be  almost no need going forward to disable rules (unless possibly any new rules brought in from future updates to `eslint-config-ash-nazg` that may be too strict to apply all at once). I have audited every rule, and to my sense, they are all reasonable, both in my view of not being too strict or too loose.

This rule in particular is important in catching security vulnerabilities. I know that this project, being SVG-based, may have a harder time doing sanitization with presumably fewer sanitizers handling SVG in addition to HTML, but even if sanitizers don't exist now, they may in the future, so I'd prefer to disable the rule inline so we can see where we are using it at least (unlike eslint config rules where their disabling can be forgotten, inline disabling of rules can more likely avoid going stale because we see them again whenever reviewing code and are thus more regularly reminded to get rid of them if possible).

Btw, until recently LGTM.com was reporting sanitization issues in a similar vein to these rules. I don't know why the alerts are [gone](https://lgtm.com/projects/g/SVG-Edit/svgedit/?mode=list) now, especially since LGTM tends to be more robust in tracing code paths, but just mentioning FYI.

Everything else looks good to me, and apparently nicely thorough too. Great work!

brettz9 (changes requested):

.eslintrc.js
jfhenon: Thanks for your review. When I migrated eslint to 7.1.0, dynamic imports such as:
````
const module = await import(`./locale/lang.${langParam}.js`);
````
were marked as unsanitized.

I don't see an issue with these imports as long as they relate to relative path to our project.

brettz9: Hmmm... I'm guessing this is because it could be problematic if `langParam` were set to `/../../different-file.js`. Of course the risk would not be as great as if it could be pointed to a remote server, but still could be something of a risk. I wonder if there is any detection of prior filtering/checks on the value of `langParam`.

jfhenon: We could check that langParam is only with alphabetical characters ('en', 'fr' et...) but I don't think this will remove the error.

jfhenon: and this would also causes an issue with userExtensions, they can be anywhere by design.

brettz9: Ok... The key thing is the source of the config. If it comes from the config file, of course that is reasonable to trust. But if can be set by URL or other untrusted input, that is where the concern lies. And it is good to continue allow setting certain params by URL, such one can point people to a bookmarkable link that preconfigures the editor properly. If the locale had allowed setting by URL, I think it is good to keep supporting that. I had already audited the URL inputs and sanitized accordingly, e.g., for `extPath` as I recall, but if you have changed the config or how they are used, yes, we should ensure that they remain secure (though for locales specifically, bear in mind that they may need to accommodate longer string formats like `pt-BR` vs. `pt-PR`; the allowable values are defined in https://tools.ietf.org/html/bcp47#page-1-5 , but I think it is enough for our purposes to just expect alphanumeric or hyphens).

brettz9: There may be the exception of the SVG not being sanitized when coming from the likes of local storage, but besides the fact that this should be safer, we have no choice I think but to trust the storage when storage is enabled, unless to strip out event listeners or such, or otherwise, we can't rebuild their SVG.

jfhenon: I changed the code so it would use encodeURIComponent on the language locales and core extensions. If anyone is trying to inject a code like "/../../different-file.js" this would be transformed into "%2F..%2F..%2Fdifferent-file.js".
In addition I relaxed the eslint rule so he allows a variable using this escape.

jfhenon: Obviously, we can't apply strong protection on the user extensions if we want our users to gave the freedom to choose their path to their extensions.

brettz9: Ok, sure. I don't remember if config is settable by URL by default, but if it is, you will want to blacklist user extensions config so it can't be. But still good to validate as you did, if nothing else than to prevent user error and avoid automated schools thinking it may come from untrusted input.

brettz9 (approved):
I think that should work. Thanks!

PR #442, opened 2020-10-25
Draft PR: move jQuery outdated UI components to Web Components

jfhenon:
## PR description

As the title says, this PR is intended to move jQuery outdated UI components to Web Components. Keeping an outdated version of jQuery is an issue from a security stand point. In addition, svg-edit is a very large file to maintain so making components is also a good direction. 

I have considered several alternatives (Vue, React, Angular and Web components). Vue (because of traction and simplicity) was a serious option. React as well was an option because there are many skills available  just because ... I know it). However, I think Web Component:
-  is a native option supported by all modern browsers (and with polyfills for older versions), 
- has a logic closer to jQuery/HTML so  we can be close to the current logic of svgEdit
- is not as advanced than other frameworks such as state management, type checking, large ecosystems etc... but it should be enough for our goal of replacing jQuery UI components.

This PR is really for "comments" at this stage. Only one component has been developed (se-button) to show the overall concept (see component folder). We will have to develop several other components: dropdown, flyout buttons, ...

The se-button component is used to replace several div used as buttons.
`````
 <se-button id="tool_source" title="Edit Source" shortcut="U" src="./images/source.svg"></se-button>
``````
The advantage is that we can group a lot of our logic (shortcut, title, image)and style at the same location: se-button.js and we can also groups several behavior in properties of index.html such as the name of the image of the button, the size, ... instead of coding this inside svgedit as today.


brettz9:
Haven't looked at the code, but a switch to Web Components sounds great to me. It might be worth looking at the issues for the various frameworks at https://custom-elements-everywhere.com/ to see whether certain features of Custom Elements might at present create some interoperability problems, but your approach SGTM.

My other minor suggestion is to avoid, where possible, something like the "se-" prefix and see whether the component can be made more generic (e.g., to `<fancy-button>` or whatever). Besides potentially allowing for reuse outside the library or contributions to our library, such thinking might inspire eventual replacements of our interim solutions for existing generic components that can be used, increasing not only modularity, but perhaps having more convenient features we could use and tapping into already well-tested components.

Re: `index.html`, I still hold out hope that we can get rid of this in favor of the opportunity to dynamically build the initial editor (including by SSR), including for the sake of optimized i18n which can render immediately into the desired language, but anything that makes the initial rendering more clean should be helpful now and toward this long term goal--as long as the long-term goal is kept in mind.

jfhenon:
see V7 preview branch

PR #446, opened 2020-11-09
Refactor canvas in multiple pieces to increase maintainability

jfhenon:
## PR description

This PR is a significant steps to increase the maintainability of svgedit by splitting svgcanvas in smaller pieces.

We need every one able interested in helping the project to participate into our effort. 
- updating the code into more recent ES6 syntax
- splitting even further svgcanvas
- adding unit/integration test
- adding a build specific to svgcanvas (currently, svgcanvas is build with svgedit) so we could push it to npm.
- ... 

## Checklist

- [ X] - Added Cypress UI tests
- [ X] - Ran `npm test`, ensuring linting passes and that Cypress UI tests keep
- [ N/A] - Added any user documentation. Though not required, this can be a big


src/common/layer.js
```diff
@@ -17,9 +17,9 @@ const $ = jQuery;
  * an existing group element or, with three parameters, will create a new layer group element.
  *
  * @example
- * const l1 = new Layer('name', group); // Use the existing group for this layer.
- * const l2 = new Layer('name', group, svgElem); // Create a new group and add it to the DOM after group.
- * const l3 = new Layer('name', null, svgElem); // Create a new group and add it to the DOM as the last layer.
+ * l1 = new Layer('name', group); // Use the existing group for this layer.
+ * l2 = new Layer('name', group, svgElem); // Create a new group and add it to the DOM after group.
+ * l3 = new Layer('name', null, svgElem); // Create a new group and add it to the DOM as the last layer.
```
brettz9, line 22: Why is this passing linting? I would expect the rule against undeclared variables should still apply. I'd prefer to keep it applying also.

brettz9 (commented):
Great to see this division work moving forward also...

Just my comment about the linting. While `@example` shouldn't need to include everything such as "use strict" and `import`, I do think it is clearer that this is indeed valid JS being offered if we declare variables. Of course for longer examples, it also telegraphs that the variable is not being overwritten.

src/common/layer.js
jfhenon: Good point. The previous code was actually not passing eslint which I found strange. However, as it was an example, I took the shortcut to remove the declaration as it was solving the issue.

jfhenon: fixed

brettz9:
Didn't do a deep review if you made any other changes besides moving the content, but in any case, leave it to you...LGTM...

PR #448, opened 2020-12-03
Issues/447 Vertical resize issue fixed

AgriyaDev5:
## Checklist


jfhenon:
close duplicate PR

PR #458, opened 2020-12-23
- Fix #450: Cancel buttons are now working

### LINKED ISSUE #450: Demo: SVG source --> "Cancel" not working
**PLEASE NOTE: This project is not currently being actively developed due to the core developers having moved on, and the only remaining maintainer merely applying occasional minor PRs or tweaks to keep the project alive. While you can file an issue, you should not expect any action, even if we label the issue.**

**Describe the bug**

I tried the demo, and if you click on the "SVG source" icon, I can see the SVG source. But the cancel button is not working. Only "Apply" closes the pop-up window.

I use Chrome on Linux.

tdittmann:
## PR description
Fixed the cancel buttons that are not working. Problem is that in svgedit.js file the selectors are used with "$q" and that only returns a single element. Therefor when multiple elements are given only one is working. I changed it so every selector has it's own object. 

## Checklist


jfhenon (approved):

jfhenon:
Thanks!

PR #465, opened 2021-01-03
V7 preview

jfhenon:
## Checklist


tdittmann:
I really look forward to this, it looks better than now :) There are some things, that are actually not working and made me think about the UI: maybe it looks tidier when the settings for an object (rect, text, ...) are placed on the right side like the layers. There are some category buttons to modify the object (e.g. Typography for texts, Meta for all objects, ...). By clicking the button the corresponding panel is opened. Here a small mockup (the button texts should be icons):
Bildschirmfoto 2021-01-13 um 11 00 39
I don't know if there already exists a modern style guide for this editor but i think that would be a great improvement

PR #466, opened 2021-01-08
Added object list to layers panel

tdittmann:
## PR description

Added a object list to the layers panel. There you can view all objects in this layer and hide/view or select them.
Bildschirmfoto 2021-01-08 um 10 10 28

## Checklist


jfhenon:
I will look into your PR this week. seems really a good addition.
Thanks
JF

jfhenon:
I can't resist to merge this excellent addition. However, please note 2 points: 
- We need cypress tests for all part of the products. 
- We will have to rewrite this functionnality in V7. 
Many thanks 
JF

PR #469, opened 2021-01-13
- Added text decoration options (underline, overline, line-through)

tdittmann:
## PR description
- Added german translations to extensions
- Added text decoration options (underline, overline, line-through)

Bildschirmfoto 2021-01-13 um 12 06 43

## Checklist


jfhenon:
Hello, 

- Can you add tests for these new buttons? (you can add to scenario.js and update snapshots)
- The dropdown is a good idea to save some space in the toolbar but it should also be used for the anchors.
- Your changes will have to be partially rewritten for V7

Thanks 

JF

tdittmann:
Added tests and changed the text anchors to a dropdown ;)

jfhenon (approved):

PR #476, opened 2021-01-21
V7 preview improvements

tdittmann:
## PR description
- Small CSS improvements
- Added panel right
- Integrated layer & object list into panel right
- Added Text Decoration Feature

## Checklist


jfhenon:
Hello @tdittmann 
I missed your PR because we actually do most of the work about V7 preview in the following fork:
https://github.com/OptimistikSAS/svgedit
I update from time to time the original project in the V7 preview branch.
It would be really good if you can work on the fork so we are sure we don't miss any of your (excellent) work.
thanks
JF

Issue #374, opened 2019-12-28
Polygon number-of-sides cannot be edited after placed.


mazerakham:
**Describe the bug**
When using the polygon tool, when you change the number of sides after you've already placed the polygon, it does not actually change the points.

**To Reproduce**
1. Create a new svg-edit document.  
2. Place a polygon with, say, 5 sides.
3. Select the polygon.
4. Change the "sides" field to, say, 6.  
5. It's still a pentagon.  The SVG file it produces says "sides:6" but there are still only five points, and the graphic is still of a pentagon (only five distinct points).
```
<polygon stroke-width="5" stroke="#000000" points="
513.993896484375,241 
363.36383056640625, 448.3245544433594 
119.63921356201172, 369.13360595703125 
119.63921356201172, 112.86638641357422 
363.36383056640625, 33.67546081542969 
513.993896484375,241 " strokeWidth="5" strokecolor="#000000" fill="none" edge="256.26722" orient="x" sides="6" shape="regularPoly" id="svg_2" cy="28" cx="81"/>
```

**Expected behavior**
You now see a hexagon (6 sides) instead of a pentagon (5 sides).  

**SVG-Edit environment (IMPORTANT)**
- File for SVG-Edit: svg-editor.html
- Version: svg-edit-2.8.1
- Protocol: https

**Desktop (please complete the following information):**
 - OS: macOS Mojave
 - Browser: Chrome
 - Version 79.0.3945.88 (Official Build) (64-bit)

Edited to clarify actual behavior.

PR #582, opened 2021-06-09
#374 Polygon number-of-sides cannot be edited after placed

AgriyaDev5:
## PR description

Polygon number-of-sides cannot be edited after placed


jfhenon (commented):
the same principle should also work for stars, no?

AgriyaDev5:
> the same principle should also work for stars, no?

yes same principle used to modified and committed.

jfhenon (approved):

Issue #584, opened 2021-06-10
Path mode regression


jfhenon:
with version 5.1.0, the select button changes when working on a path as below.

https://user-images.githubusercontent.com/20402845/121500493-a4d0c480-c9de-11eb-84ed-3f4adb7b5f68.mov

This does not work on v7.

AgriyaDev5:
https://user-images.githubusercontent.com/44435283/125234783-026a7080-e2ff-11eb-99c3-52829b632a94.mp4

PR #585, opened 2021-06-10
#584 the select button changes when working on a path

AgriyaDev5:
## PR description

 the select button changes when working on a path

### DIFF
--- a/src/editor/panels/TopPanel.js
+++ b/src/editor/panels/TopPanel.js
@@ -968,7 +968,7 @@ class TopPanel {
        </div> <!-- a_panel -->
        <div id="path_node_panel">
          <div class="tool_sep"></div>
-        <se-button id="tool_node_link" title="${i18next.t('tools.node_link')}" src="./images/tool_node_link.svg" pressed="${true}">
+        <se-button id="tool_node_link" title="${i18next.t('tools.node_link')}" src="./images/tool_node_link.svg" pressed>
          </se-button>
          <div class="tool_sep"></div>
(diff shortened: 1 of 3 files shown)


src/editor/panels/TopPanel.js
```diff
@@ -968,7 +968,7 @@ class TopPanel {
        </div> <!-- a_panel -->
        <div id="path_node_panel">
          <div class="tool_sep"></div>
-         <se-button id="tool_node_link" title="${i18next.t('tools.node_link')}" src="./images/tool_node_link.svg" pressed>
+        <se-button id="tool_node_link" title="${i18next.t('tools.node_link')}" src="./images/tool_node_link.svg" pressed={${true}}>
```
jfhenon, line 971: pressed is equivalent to pressed={$(true)} ??

jfhenon (changes requested):
see my question but more generally, I didn't see the change (I used the deploy preview:
https://deploy-preview-585--svgedit.netlify.app/editor/index.html)
Another point to check is that when a path is selected and you click on it, you also go in the 'path' mode. You can see the icon changing to: 
image

src/editor/panels/TopPanel.js
AgriyaDev5:
```
<div id="tool_node_link" class="tool_button push_button_pressed" title="Link Control Points"></div>
```

Initially `push_button_pressed` class added in  `tool_node_link` button, that why pressed set true.

icon also changed.

jfhenon (approved):

Issue #475, opened 2021-01-18
Can't log into IAN library


idavillc:
When you click on the image library button and choose the IAN library. The login form does not log you into your IAN account. It just refreshes the page. I tried locally and the demos listed here at github.

I double checked the login credentials and they are correct.

**SVG-Edit environment (IMPORTANT)**

- File for SVG-Edit: ../dist/editor/index.html
- Version: 5.1.0
- Protocol: https
- Browser: Version 87.0.4280.141 (Official Build) (64-bit)

AgriyaDev5:
> Can't log into IAN library

IAN as they changed the API, so we removed IAN option in SVGEdit.

PR #610, opened 2021-08-16
#475  IAN as they changed the API, so removed IAN option in svg edit

AgriyaDev5:
## PR description

IAN as they changed the API, so removed IAN option in svg edit. The current extension is actually causing other issues. As SVG-Edit allow to import an image. I think it's not a big issue to choose and download an image on internet and to import it  after in SVG-Edit.

This extension is however needed for users that want to access quickly to a list of SVG that are related to an activity.


q464pp:
可以拉取合併嗎 謝謝

jfhenon:
I need to review this one before merging it.

q464pp (commented):
審查什麼東西?

jfhenon:
I didn't test if it is working correctly. 
I can put you as an approver if you have tested it.

q464pp:
恩 謝謝

Issue #612, opened 2021-08-17
Bottom toolbar has some visual issues 


AgriyaDev5:
**Describe the bug**
Bottom toolbar has some visual issues 

**Screenshots**
![image](https://user-images.githubusercontent.com/44435283/129686604-2e825c81-d8c9-4120-92a8-d2aabedcb7c1.png)

- [x]  opacity alignment 
- [x]  Left and bottom scroll bar style changes 
- [x]  Shape popup menu position problem
- [x]  Menu popup position issue

PR #613, opened 2021-08-17
#612 Bottom toolbar has some visual issues

AgriyaDev5:
## PR description

Bottom toolbar has some visual issues


src/editor/components/seSpinInput.js
```diff
@@ -83,6 +84,7 @@ export class SESpinInput extends HTMLElement {
     case 'src':
       this.$img.setAttribute('src', newValue);
       this.$label.remove();
+      this.$div.style.width = '82px';
```
jfhenon, line 87: I dont think we should hard code the width. We could instead
- put the width as an attribute of the web component
- define a global css variable --se-spin-input-width 
- ...

jfhenon (changes requested):
see comment please.

src/editor/components/seSpinInput.js
AgriyaDev5: modified

src/editor/components/seSpinInput.js
```diff
@@ -8,6 +8,10 @@ template.innerHTML = `
     margin: 5px 1px;
     padding: 3px;
   }
+  div.imginside {
+    --se-spin-input-width: var(--global-se-spin-input-width);
+    width: var(--se-spin-input-width);
```
jfhenon, line 13: why do we need --se-spin-input-width and --global-se-spin-input-width ?? 
Also, you added the width parameter to the se-spin-input. 
I think we need to decide between defining a global variable and reuse it inside the component or adding a width attribute to se-spin-input.

AgriyaDev5: shall i use global variable or width attribute ?

jfhenon (approved):

Issue #639, opened 2021-09-02
path #1 draws its control point handles connected to path #2


bluetwelve:
Create a path shape I used closed paths or converted circle to path). Double click this path - verify that it's control paths are correct. Create another path. Double click this path to display it's control points - verify that they are correct.

Now select the first path and expose its control points, The handles will incorrectly connect to the other shapes control points.

Note that even if you open a new image, these old control points linger and a new path's control handles will connect to the coordinates where the path used to be

Reproduces on v7 only

PR #640, opened 2021-09-03
#639 path #1 draws its control point handles connected to path #2

bluetwelve:
Fixes  https://github.com/SVG-Edit/svgedit/issues/639

pathMethodsContext_.setPathObj(this) was only being called on path instantiation. So if you edit path1, then path2 and then switch back to path1 and go into edit mode, the context is pointing at the last path you edited (path2) and the control handles for path1 are drawn connected to path2.


jfhenon:
Fast! 
I just tested your PR (for info, netlify deploys all PR for preview: https://deploy-preview-640--svgedit.netlify.app/editor/) and the issue disappeared. 
I'm waiting for the tests to pass and will merge.

jfhenon (approved):

PR #658, opened 2021-10-29
add y1 to TopPanel attributes change listener

phcreery:
UNTESTED yet.

Should fix issue when drawing line and changing the y1 value in the top panel not adjusting properly.


jfhenon (approved):

jfhenon:
> UNTESTED yet.
> 
> Should fix issue when drawing line and changing the y1 value in the top panel not adjusting properly.

 I adjusted tests (needed to update snapshots) and I also used your PR to update dependencies.
Thanks !!

PR #667, opened 2021-11-22
Fix : fix mouse-wheel javascript error

cg-scorpio:
## PR description

avoid javascript error in order to fix mouse wheel event problem

## Checklist

- [  - ] - Added Cypress UI tests
- [ x ] - Ran `npm test`, ensuring linting passes and that Cypress UI tests keep
- [ - ] - Added any user documentation. Though not required, this can be a big

### DIFF
--- a/src/svgcanvas/event.js
+++ b/src/svgcanvas/event.js
@@ -1295,7 +1295,7 @@ export const DOMMouseScrollEvent = function (e) {
   if (!e.shiftKey) { return; }
 
   e.preventDefault();
-  const evt = e.originalEvent;
+  const evt = e;
 
   eventContext_.setRootSctm($id('svgcontent').querySelector('g').getScreenCTM().inverse());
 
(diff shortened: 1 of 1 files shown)


src/svgcanvas/event.js
```diff
@@ -1295,7 +1295,7 @@ export const DOMMouseScrollEvent = function (e) {
   if (!e.shiftKey) { return; }
 
   e.preventDefault();
-  const evt = e.originalEvent;
+  const evt = e;
```
jfhenon, line 1298: Why do we need 2 variables ( e and evt) ?

cg-scorpio: We don't need tow variables. I propose this "simple" change only because I don't understand why the original code use the field "originalEvent" which doesn't exists.
If you prefer, I will simplify and use "e" everywhere.

cg-scorpio: Done here : https://github.com/SVG-Edit/svgedit/pull/667/commits/ee8b2c6296164f0f6c1a150d99e7e3a144d73759

jfhenon (approved):

PR #737, opened 2022-03-10
Extensions

jfhenon:


jfhenon:
- add the current document title in the toolbar
- allow user extensions to define optional parameters
- more events for renamedElement, beforeClear, afterClear, sourceChanged
- remove "message" event used for iframe integration not used anymore. can be readded through a new extension if necessary
- default precision down to 2 digits (and applied in rotation)

PR #755, opened 2022-04-22
Added tr language and translation

zktosu:
## PR description

added tr file for localization and translated 80% or so. This should be enought for initial use cases.

## Checklist


zktosu:
fixed the related error ( missing a comma on line 165: lang.tr.js)

jfhenon (approved):

jfhenon:
it seems the linter stii does not pass. can you please check?

PR #822, opened 2022-07-23
Enable text tool panel wen multiple text elements are selected.

pmkrawczyk:
## PR description

This PR enables text tool panel wen multiple text elements are selected. This will be extended to other element types in subsequent PR.

## Checklist

### DIFF
--- a/src/editor/panels/TopPanel.js
+++ b/src/editor/panels/TopPanel.js
@@ -422,20 +422,10 @@ class TopPanel {
 
       // if (elem)
     } else if (this.multiselected) {
-      // Check if all selected elements have the same tag
+      // Check if all selected elements are 'text' nodes, if yes enable text panel
       const selElems = this.editor.svgCanvas.getSelectedElements()
-      const tagNames = []
-      selElems.forEach(elem => {
-        const { tagName } = elem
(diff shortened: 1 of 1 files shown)


src/editor/panels/TopPanel.js
```diff
@@ -422,6 +422,22 @@ class TopPanel {
 
       // if (elem)
     } else if (this.multiselected) {
+      // Check if all selected elements have the same tag
+      const selElems = this.editor.svgCanvas.getSelectedElements()
+      const tagNames = []
+      selElems.forEach(elem => {
```
jfhenon, line 428: while it may not be a performance or memory issue, do we have:
1. keep the tag names in the array tagNames?
2. do this loop even if tagNames[0] is not 'text"

I would suggest to use every() on selElems and to check if they are all 'text'. 
every would stop on the first tag not 'text'

jfhenon (changes requested):
proposed improvements - see comments

src/editor/panels/TopPanel.js
pmkrawczyk: Thanks for the suggestion, much cleaner this way!

jfhenon (approved):

Issue #671, opened 2021-12-01
Rewrite the zoom selector


benjaminbch:
**NOTE:**
This issue is now funded on Bountysource (https://app.bountysource.com/issues/109641308-minor-display-issue-with-zoom-selector)
However, the fix needs to be a **FULL REWRITE** of the zoom selector web component without using the elix library. 
The component source is: src/editor/components/seZoom.js
Native JS, CSS preferred but another library is possible if it is popular and actively maintained.

Obviously, the fix needs to work consistently on Safari, Firefox and Chrome.

**To Reproduce**
1. Open zoom selector

**Expected behavior**
All options must be visible with correct hover

**SVG-Edit environment (IMPORTANT)**
- Version: 7.0.0

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser : Firefox
 - Version : 91 ESR

**Screenshots**
![unnamed](https://user-images.githubusercontent.com/63210012/144289018-8d260aaa-0922-415a-973c-f10d1f1aa98c.png)

luciozambon:
The same bug with SVG-Edit environment 
    Version: 7.0.0

Desktop:
    OS: Windows 10
    Browser : Google Chrome
    Version : 97

This bug isn't present on MacOS, both Chrome and Safari

pratik2315:
Hey! Anyone working on this issue? Im interested

jfhenon:
> Hey! Anyone working on this issue? Im interested

I don't think so. You can start on this issue. 
JF

radasam:
Hi, 

What's the intended styling for the popup part of the selector? 

I assume at least the hover styling is broken in the screenshot and latest build

Thanks

jfhenon:
We have no strict guidelines but clearly the current style is indeed broken.

jfhenon:
@radasam you can claim your payment on bountysource.
JF

radasam:
@jfhenon thanks, I've signed up and submitted a solution

PR #826, opened 2022-08-05
Fix: Zoom selector rewrite

radasam:
## PR description

A full rewrite of the zoom selector in native js/css fixes issue #671 

Added tests for the zoom tool, under /cypress/e2e/unit/zoom.cy.js let me know if there is a more appropriate place for these.

## Checklist


jfhenon:
I tested your work (https://deploy-preview-826--svgedit.netlify.app/editor/) and it's promising. 
Please fix a few things:
- Linter issues (we use standardjs) - use `npm run lint` and `npm run test `if you want to test your PR locally before pushing it to github.
- When you enter in the input field, you cannot type directly a number (I tried with 35...and you get 3.0)
- The list of choices should finish with % (100 -> 100%)
- The CSS for the background of the input field should be consistent with other (light gray)
image

jfhenon (changes requested):
see discussions

jfhenon:
I fixed the linter issues as well as the CSS.
Can you look at the remaining questions?

radasam:
- Fixed the input issue - I've changed the behaviour so we now wait for the user to click off the input before triggering a zoom change.
- Suffixed each of the numerical values in the popup with %
- I've also changed the css slightly to align with the style of the other tools on the bottom panel

jfhenon (approved):

PR #827, opened 2022-08-06
refactor se-list and se-list-item

### LINKED ISSUE #784: Rewrite the seList and seListItem components
NOTE:
This issue is funded on Bountysource (https://app.bountysource.com/issues/109641309-rewrite-the-selist-and-selistitem-components)
However, this is a FULL REWRITE of the list/listitem selector web component without using the elix library.
The component source is: src/editor/components/seList.js (and seListItem.js)
Native JS, CSS preferred but another library is possible if it is popular and actively maintained.

Obviously, the fix needs to work consistently on Safari, Firefox and Chrome.

ThakurKarthik:
## PR description

Refactor code for couple of components using plain javascript instead of elix library

closes #784 

## Checklist


jfhenon (changes requested):
I tested your PR with https://deploy-preview-827--svgedit.netlify.app/editor/
This seems to work well but we need to fix a few things before merging:
- linter issues (please make sure you run locally `npm run lint` and `npm run test` before pushing your changes. 
- The toolbar is spreaded on 3 lines in the top bar when you  select a line: 
 
image

- The selected item needs is marked with a black border like below: 

image
The selected class should be improved (look at the .pressed class of the left bar ): 
image

`.selected {
    background-color: var(--icon-bg-color-hover);
}`
```

File: src/editor/EditorStartup.js [hunk 1]
```diff
@@ -314,6 +314,16 @@ class EditorStartup {
       return true
     })
 
+    // adding a mousedown listener to handle some dropdowns closing
```
jfhenon, line 317: could this mousedown event be trapped inside the selist component that would close itself?
The idea is to avoid spreading logic in multiple place.

jfhenon (commented):
Can you answer to my question in the review?
This is the only remaining issue according to me.
Thanks
JF

jfhenon (approved):

PR #845, opened 2022-10-30
Keep aspect ratio when resizing objects with 'preserveAspectRatio' attr

pmkrawczyk:
## PR description

This PR adds a check to svgcanvas/event.js to figure out whether the resized object has 'preserveAspectRatio' set. If yes, the aspect ratio of the resized object will be maintained as if the user held shift. This is useful when external objects (e.g. another SVG image) are added where aspect ratio should be preserved.

Since this code now lives in a separate package, I'm not 100% sure whether I have tested it properly.

## Checklist

- [] - Added Cypress UI tests

### DIFF
--- a/packages/svgcanvas/event.js
+++ b/packages/svgcanvas/event.js
@@ -141,6 +141,9 @@ const mouseMoveEvent = (evt) => {
     y = snapToGrid(y)
   }
 
+  const aspectRatioAttr = selected.getAttribute('preserveAspectRatio').split(' ')
+  const preserveAspectRatio = ['xMinYMin', 'xMidYMin', 'xMaxYMin', 'xMinYMid', 'xMidYMid', 'xMaxYMid', 'xMinYMax', 'xMidYMax', 'xMaxYMax'].includes(aspectRatioAttr[0])
+  
   let tlist
   switch (svgCanvas.getCurrentMode()) {
     case 'select': {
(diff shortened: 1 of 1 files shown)


jfhenon:
The **preserveAspectRatio** defines the behavior of the SVG when in a viewBox. It's not really a way to define how it behaves inside an editor. A user may want to change the size/ratio of a SVG even if this SVG has the preserveAspectRatio defined. 
In the current version, you can press the shift key to preserve the ratio when you resize the SVG, isn't it sufficient for your use case?

pmkrawczyk:
I agree, it is not a very clean solution. But the current behavior of the editor when resizing fixed-ratio content is very confusing for users. For instance, when user changes height only, the resize box and handles are resized but content remains same size, with white space added between the content and resize box. And the user has no idea why this happens. I tried, but cannot find a scenario where this is a desired behavior. And unfortunately while shift is a perfect solution to this, non-advanced users have no idea that this is possible.

Althought this is not relevant here, one could even make a strong case for reversing the shift modifier bahavior specifically for images, so that they always maintain aspect ratio, except when shift is held. This is default in Powerpoint for instance, and for a good reason - one very rarely needs to change aspect ratio of images.

Anyway, if my PR is not fitting, would you have any suggestions as to how to achieve the fixed aspect effect via a plugin for instance?

jfhenon:
I'm not against reversing the default behavior with the shift button.

pmkrawczyk:
Great, I will open another PR - thank you for the productive exchange! :-)

Issue #853, opened 2022-11-23
text element font-family select in top panel does not update to selected value


bluetwelve:
Create three text items in an svg. Use the top panel to select a different font-family for each. Now select each of the three in turn.

Expected behavior: the correct font-family for each text element will be displayed in the top panel

Actual behavior: the displayed font-family is always the last one selected when you were changing the elements

PR #856, opened 2022-11-28
fix bug #853 -  font family select doesn't update the Top Panel dropdown

bluetwelve:
Fixes https://github.com/SVG-Edit/svgedit/issues/853#issue-1462248691


jfhenon:
good catch! 
We have issues with netlify that I need to fix before merging your fix.
Thanks
JF

PR #917, opened 2023-08-31
Fix rimraf import

pmkrawczyk:
## PR description

This fixes rimraf import from:
import { rimraf } from rimraf

To:
import pkg from 'rimraf';
const { rimraf } = pkg;

The current import is incorrect and breaks in certain environments:
[!] SyntaxError: Named export 'rimraf' not found. The requested module 'rimraf' is a CommonJS module, which may not support all module.exports as named exports.

## Checklist


pmkrawczyk:
The issue was caused by our CI infrastructure, closing!

Issue #918, opened 2023-09-14
Actions Not Working When Selecting Multiple Text Elements


Moath-Zaghdad:
**Bug Description**

Several actions are not functioning as expected when selecting multiple text elements.
The affected actions include Bold Text, Italic Text, Underline, Line-Through, Overline, Text Alignment, Letter Spacing, Word Spacing, and possibly others.

![CleanShot 2023-09-14 at 11 08 03@2x](https://github.com/SVG-Edit/svgedit/assets/18503873/0d357f5d-b9f5-4b10-a05c-c3fba16a9295)

**To Reproduce**
To replicate the issue, follow these steps:
1. Visit the SVGEdit tool at https://svgedit.netlify.app/editor/index.html.
2. Create two Text elements.
3. Select both Text elements.
4. Attempt to use any of the aforementioned actions.

Expected Result: The selected actions should apply to both text elements.
Actual Result: None of the actions work when selecting more than one text element.

**SVG-Edit environment**
- Version: 7.3.2
- Protocol: https

**Desktop Information:**
 - OS: iOS
 - Browser: safari
 - Version: 16.6

Moath-Zaghdad:
Guidance on resolving this issue would be greatly appreciated, and I am eager to contribute to its resolution.

jfhenon:
Thank you for detecting this issue and for your proposal to fix thix. 

I think the fix is to change the function setBoldMethod in elem-get-set.js

This function works only on the first selected element: 

```
/**
* Make the selected element bold or normal.
* @function module:svgcanvas.SvgCanvas#setBold
* @param {boolean} b - Indicates bold (`true`) or normal (`false`)
* @returns {void}
*/
const setBoldMethod = (b) => {
  const selectedElements = svgCanvas.getSelectedElements()
  const selected = selectedElements[0]
  if (selected?.tagName === 'text' &&
    !selectedElements[1]) {
    svgCanvas.changeSelectedAttribute('font-weight', b ? 'bold' : 'normal')
  }
  if (!selectedElements[0].textContent) {
    svgCanvas.textActions.setCursor()
  }
}
```

but it should indeed iterate through all selected element so it should look more like:
```
/**
* Make the selected element(s) bold or normal.
* @function module:svgcanvas.SvgCanvas#setBold
* @param {boolean} b - Indicates bold (`true`) or normal (`false`)
* @returns {void}
*/
const setBoldMethod = (b) => {
[...]

PR #919, opened 2023-09-18
fix: Enhance text properties for multiple selected elements (SVG-Edit/svgedit#918)

Moath-Zaghdad:
This commit addresses issue SVG-Edit/svgedit#918 by improving text property handling when multiple elements are selected. You can now set Bold, Italic, Text-Decoration, Text-Anchor, Letter-Spacing, Word-Spacing, Text Length, and Length Adjust.


jfhenon:
The update of the top toolbar works inconsistently based on the first element in selectedElements. However, this is a useful PR so I merge it.

image

PR #937, opened 2023-12-23
Bottom panel styling, more adaptive palette, shortcuts

olekhshch:
- vertical scrollbar on the right of the bottom panel (BP) removed
- changed coordinates of the colour picker windows so they appear fully on the screen (previously, those windows were overflowing the editor's container)
image

- the palette is more adaptive to the screenwidth (shrinking, showing less colours if doesn't fit) to prevent earlier scrollbar appearance
- if necessary, the palette can be expanded (arrow button to the right), so the whole palette is available without scrolling the BP. The popup can be closed by clicking the same button or anywhere on the svgcanvas
image
- BP's scrollbar's styling is the same as the styling of other toolbars
image

- preventDefault was added to the selectAll shortcut to avoid selecting the whole editor content (as on screenshot) (usually this selection appears after clicking a button on any panel and selecting svg objects with ctrl+A after that)
image

- also, preventDefault was added to rotation shortcuts (crtl+arrowLeft) because some browsers (e.g. OperaGX) are using this shortcut for navigation through routing history.

P.S. My Prettier configurations added some linebreaks to the edited files, but the `npm run lint` didn't catch it as an error, please let me know if I still should undo it.

### DIFF
--- a/.github/workflows/onpullrequest.yml
+++ b/.github/workflows/onpullrequest.yml
@@ -1,6 +1,9 @@
 name: Node CI
 
 on: [pull_request]
+# To be able to write in the PR, we need write permission. 
+permissions:
+   pull-requests: write
 
 jobs:
   build:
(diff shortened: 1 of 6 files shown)


jfhenon (approved):

jfhenon:
Thanks for the PR

PR #939, opened 2023-12-27
Panning, modeChange Event, minor changes

olekhshch:
-- prDefault for pressing on `mouse wheel` to avoid firing of default browser's (e.g. Chrome) panning

-- quick access to the _select_ mode while _panning_ mode is active with dbClick. Also, the _panning_ mode is active while holding `Space/mousewheel` - once they're released the mode changes back to the previous one. This allows a quicker use of both Select and Panning tools without a need to click them on the left panel.

-- new custom event was added in svgcanvas that fires on every setMode(). For now, the event helps to track the active mode to set cursor styling (atm only zoom, panning, and select tools have their own cursors, but I would like to add it for other tools for better UX). Also, this event can have different implementations in the future.

-- prDefault for `Ctrl+wheel` - I've added it based on my own experience because I always try to scale the canvas with this shortcut (instead of `Shift+wheel`), but it scales the whole page instead. Having this behavior here seems unnecessary, perhaps it would be better to allow this shortcut to scale the canvas as well in the future. 👀 

-- `Ctrl` was added to the Undo/Redo shortcuts because the` Z` shortcut was already assigned for the _Zoom_ tool.

### DIFF
--- a/packages/svgcanvas/svgcanvas.js
+++ b/packages/svgcanvas/svgcanvas.js
@@ -308,6 +308,9 @@ class SvgCanvas {
     this.contentW = this.getResolution().w
     this.contentH = this.getResolution().h
     this.clear()
+
+    // creates custom modeEvent for editor
+    this.modeChangeEvent()
   } // End constructor
 
   getSvgOption () {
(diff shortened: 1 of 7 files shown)


jfhenon (approved):

jfhenon:
The build breaks because of the NYC coverage. The tests are working. Thank you for this PR.

Issue #896, opened 2023-03-29
extensions `zoomChanged` action is not being called


Moath-Zaghdad:
The grid is not showing the correct grid scale on Zoom change
I tried tracking the issue, and I've noticed that the extension's [zoomChange](https://github.com/SVG-Edit/svgedit/blob/f47759867af79031e5465d660cf7f90314ac4a3f/src/editor/extensions/ext-grid/ext-grid.js#L157) method is not being called.

**To Reproduce**
1. Open [SVG-Edit V7.3.0](https://unpkg.com/svgedit@7.3.0/dist/editor/index.html) (or master)
2. Change the zoom to `50%`
3. Click on the `Show/Hide Grid` button
4. Change the zoom to `1000%`
5. Click on the `Show/Hide Grid` button twice to see what should have happened automatically on zoom change

**SVG-Edit environment**
- Version: 7.2.0

**Desktop:**
 - OS: iOS
 - Browser: Firefox Developer Edition
 - Version: 112.0b7 (64-bit)

jfhenon:
I'm not able to reproduce the issue. Here is the video of the test I did following your instructions:

https://user-images.githubusercontent.com/20402845/228971790-5c30861d-ccd4-4a5b-80f5-0ca8f7af1ea8.mov

Moath-Zaghdad:
Sorry, the reproduce steps were missing one thing, start with a zoom of 50%

https://user-images.githubusercontent.com/18503873/229141766-b7cf19e3-4d2c-4b29-8833-c8755c5cb70f.mp4

jfhenon:
I could reproduce. I'm not yet sure where to look for but the changeZoom function is in BottomPanel.js

Moath-Zaghdad:
I have created this PR https://github.com/SVG-Edit/svgedit/pull/940 to fix this issue

PR #940, opened 2023-12-29
fix: Call zoomChanged event after updating the zoom

Moath-Zaghdad:
The zoomChange method was not being called on the extensions, and this commit fixes this by calling the runExtensions method after updating the zoom.
This commit addresses issue SVG-Edit/svgedit#896


jfhenon (approved):

jfhenon:
thanks

PR #946, opened 2024-01-15
Cursor

olekhshch:
## PR description

-- more specific cursors for _Circle_, _Ellipse, Square_ and _Rectangle_ tools

-- `Esc` key cancels the currently selected tool. It applies only for _Zoom, Rect, Square, Circle, Ellipse, Line, Text, Star, Polygon,_ and _Eyedropper_. Cancelation with the `Esc` key is disabled if pressed in the middle of using those tools (e.g. in the middle of drawing a new circle) to avoid unexpected behaviour;

-- `I` and `B` shortcuts for _italic_ and _bold_ text are unavailable if no text element is selected - currently, when pressing those keys nothing happens and other shortcuts (e.g. E for Ellipse) become unavailable until refresh (app throws an error);

-- minor styling changes

PR #947, opened 2024-01-20
fix #832, floating btns closing, Esc to close confirmation dialog, 0x0 rects disabled

### LINKED ISSUE #832: How to keep the mode after I draw a line circle etc.
Is there a way to keep the mode not change to select mode after a draw a line

olekhshch:
## PR description

-- `Esc` key cancels the dialog window appearing before creating a new document (N) or opening new svg file (atm Esc key acts as Enter and selects currently focused button (usually it's an "Ok" choice) which is a bit confusing);

-- Floating buttons (e.g. _Shape Library_ or _Rect/Square/Rect by hand_) can be closed by clicking somewhere in the document (_shapeLibrary_ - by clicking on the workarea and button itself, not only on the handle);

-- _ShapeLibrary_ can be cancelled with Esc key (same as [here](https://github.com/SVG-Edit/svgedit/pull/946));

-- Drawing of 0x0 rectangles/squares is disabled (atm a simple click while in _rect_ or _square_ creates such figure)

-- Addressing #832 - the mode will not change to _select_ if `Alt` was pressed. I've chosen this key because Shift is reserved for Circle/Square while drawing Ellipses/Rect.

-- after drawing a shape from the _Shape Library_ (if `Alt` is not pressed) the mode changes to _'select'_ (same as with other tools, atm after drawing a shape mode doesn't change automatically)

### DIFF
--- a/packages/svgcanvas/core/event.js
+++ b/packages/svgcanvas/core/event.js
@@ -713,7 +713,9 @@ const mouseUpEvent = (evt) => {
       const width = element.getAttribute('width')
       const height = element.getAttribute('height')
       // Image should be kept regardless of size (use inherit dimensions later)
-      keep = (width || height) || svgCanvas.getCurrentMode() === 'image'
+      const widthNum = Number(width)
+      const heightNum = Number(height)
+      keep = widthNum >= 1 || heightNum >= 1 || svgCanvas.getCurrentMode() === 'image'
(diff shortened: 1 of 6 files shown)


jfhenon (approved):

jfhenon:
Thanks

PR #966, opened 2024-04-25
Update lang.ru.js

KamilOcean:
I'm a native Russian speaker, I've added some new lines

## Checklist


src/editor/locale/lang.ru.js
```diff
@@ -216,12 +216,12 @@ export default {
     snapping_stepsize: 'Шаг привязки:',
     grid_color: 'Цвет сетки',
     done: 'Done',
-    change_xxx_color: 'Change xxx color',
+    change_xxx_color: 'Поменять xxx цвет',
```
jfhenon, line 219: Does not sound as Russian?

KamilOcean: It's Russian. What does `xxx` mean? I thought it's like a template, isn't it?

jfhenon: My mistake. Thank you for the contribution

jfhenon (approved):

PR #972, opened 2024-05-25
fix double binding of shortcut "d"

spaghetti22:
## PR description

Hi, I noticed that the shortcut "d" was bound to both "document properties" and "duplicate" and managed to fix this. I hope I am doing this right, this is the first time I am really doing a github pull request on my own.

I did not add a UI test because I did not see that any other shortcuts were currently tested. If I need to do that anyway, please tell me and I will do my best.


jfhenon (approved):

jfhenon:
Thank you

PR #993, opened 2024-09-16
Populate layers after new file is opened

VenissaCarolQuadros:
## PR description
With this fix the layers panel is updated when a new file is opened. Earlier, when an SVG file was opened the Layers panel was not populated with the correct number of layers and instead only showed 'Layer 1' irrespective of how many layers the newly opened file contained. This is fixed with a single-line change within the 'opensave' extension. 

## Checklist


jfhenon (approved):

VenissaCarolQuadros:
Thanks for the quick review!
Just a reminder that this is still awaiting merge from someone with write access.

PR #995, opened 2024-09-26
Merge layer_view extension

VenissaCarolQuadros:
## PR description
Created a new extension called 'layer_view' extension in the TopPanel, which can easily make only the current layer visible. I find myself repeatedly doing but have to do this except it earlier required clicking the eye icon on each of the layers which gets tedious if there are many layers. 

It has been designed as follows:
1. When turned on the 'layerinvis' is included in the class list of all but the current layer
2. The tool can be disabled either by clicking on the tool again or clicking on one of the 'eye' icons on the side. The latter is done to not disrupt the functionality of the LayerPanel.

I realize this is not a requested feature but thought it would be a neat feature to add. The default state in the ConfigObj.js is set to false but could be set to true id required. 

## Checklist


jfhenon (approved):

jfhenon (changes requested):
image
The button is not styled like others (white/orange). Could you fix this? 
The rest is good and thank you for the addition.

PR #1028, opened 2025-03-02
Fixed issue in EditorStartup.init() where $qq will not work if the editor is in a Web Component

mscherotter:
## PR description

in EditorStartup init(), 
` this.$svgEditor = $qq('.svg_editor')` 
will return undefined if the editor is in a web component, Changing the line to 
`this.$svgEditor = this.$container.querySelector('.svg_editor')` 
will make the code work whether it is in a web component or not.

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
       "dependencies": {
         "@svgedit/svgcanvas": "7.2.3",
         "browser-fs-access": "0.35.0",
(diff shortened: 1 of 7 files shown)


jfhenon (approved):

jfhenon (commented):
need to check linter issue

jfhenon (approved):

PR #1040, opened 2025-03-25
Fixed bug in Editor.loadFromURL where the promise was not resolving

mscherotter:
## PR description

on successful load the Editor.loadFromURL() function returns the string, instead of resolving the string in the promise.
The result of this bug is that the function never returns.  This bug fixes that.

## Checklist

- [X] - Added Cypress UI tests

## Summary by Sourcery

Fixes a bug where the Editor.loadFromURL() function was not resolving the promise on successful load, causing the function to never return. The fix ensures that the promise resolves with the loaded string.

Bug Fixes:
- Fixes a bug where the Editor.loadFromURL() function was not resolving the promise on successful load.

Tests:
- Adds Cypress UI tests to ensure the fix is not regressed in the future.

### DIFF
--- a/src/editor/Editor.js
+++ b/src/editor/Editor.js
@@ -1353,7 +1353,7 @@ class Editor extends EditorStartup {
           })
           .then((str) => {
             this.loadSvgString(str, { noAlert })
-            return str
+            resolve(str);
           })
           .catch((error) => {
             if (noAlert) {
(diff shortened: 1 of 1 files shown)


jfhenon:
Please remove the semi column so the tests pass.

jfhenon (approved):

jfhenon:
👍

PR #1044, opened 2025-04-06
Added uppercase HTML elements to sanitize to as lowercase HTML elment…

mscherotter:
This fixes an issue where the SVGCanvas was converting lowercase foreignObject HTML tags to capitalized ones.

## Checklist

## Summary by Sourcery

Bug Fixes:
- Fix issue where SVGCanvas was only allowing lowercase HTML tags in foreignObject, now supporting both lowercase and uppercase variants

### DIFF
--- a/packages/svgcanvas/core/sanitize.js
+++ b/packages/svgcanvas/core/sanitize.js
@@ -109,20 +109,35 @@ const svgWhiteList_ = {
 
   // HTML Elements for use in a foreignObject
   div: [],
+  DIV: [],
   p: [],
+  P: [],
   li: [],
+  LI: [],
   pre: [],
+  PRE: [],
   ol: [],
+  OL: [],
   ul: [],
+  UL: [],
   span: [],
+  SPAN: [],
   hr: [],
+  HR: [],
   br: [],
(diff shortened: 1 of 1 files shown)


jfhenon:
Ins't it better to consider the comment of Sourcery? Instead of adding uppercase versions of the tags, consider converting the tag names to lowercase before checking against the whitelist

mscherotter:
@jfhenon there is a bug that I discovered in the code called in showSourceEditor() which calls this.editor.svgCanvas.getSvgString() that generates the SVG xml.  In the process of generating that XML, all of the HTML elements in the foreignObject are converted to uppercase.  I haven't figured out why, but this fix will mitigate that issue. Since many of the non-HTML elements in the whitelist are camelCase, the suggested method will not work.

mscherotter:
Please accept this pull request until we can figure out how to fix the issue with getSvgString() that converts HTML elements to upper-case.

jfhenon (changes requested):
I'd prefer a change that fix the issue including with getSgvString

mscherotter:
Root issue was fixed

PR #1046, opened 2025-04-08
Fix SVG files to have a separating space between the comments and URLs

unique-EJ:
## PR description

Fix SVG files to have separating space between the comments ("Created with SVG-edit - ") and URLs - this lets people use the URLs by double-click without the end comment being added, this was fixed in other files - src/editor/images/hello_world.svg

## Checklist

I edited through GitHub and the added changes are a space inside a single comment element... as the tests must be run, would someone _help please_ (to do this).

Documentation would be my commit message.

## Summary by Sourcery

Bug Fixes:
- Modify SVG metadata comments to include a space between the 'Created with SVG-edit' text and the URL, allowing easier URL selection by double-clicking


unique-EJ:
I think there is an issue in packages/svgcanvas/core/svg-exec.js with the space between the comment and the URL, this PR did not fix that.

Please fix the issue, if I open the new image _untitled.svg_ in the _SVG-edit_ Deploy Preview and then edit source (shortcut: <kbd>U</kbd>), there is _not_ this change.

jfhenon (changes requested):
This is only fixing existing images. needs to fix as well where this mention is created.

PR #1053, opened 2025-05-05
Fixed issue where foreignObject contents styles were getting removed on object moving

mscherotter:
## PR description

when moving foreignObject elements, the styles in the contents were getting removed.  This fixes that issue so that style attributes in the contents of foreignObjects do not get removed on a move.

## Checklist

## Summary by Sourcery

Preserve styles for foreignObject contents when moving SVG elements

Bug Fixes:
- Fixed an issue where styles were being removed from foreignObject contents during element movement

Enhancements:
- Added special handling for foreignObject elements to preserve their internal styles


mscherotter:
@jfhenon  - ping again, can you please update the NPM Package with the latest fixes?

PR #1083, opened 2026-03-15
Fix/rotation recalc compound transforms

geoffyoungs:
## PR description

Fix rotation centre recalculation corrupting compound transforms

When changing attributes on rotated elements, the rotation center recalculation had two bugs:

1. It fired for all attribute changes (stroke-width, fill, opacity, etc.), not just geometry-affecting ones (x, y, width, height, etc.). Non-geometric changes don't move the bbox centre, so the recalculation was unnecessary and destructive.
2. The new rotation center was computed through all remaining transforms in the list, including pre-rotation ones like translate. This caused the translate to leak into the centre, producing incorrect rotation.

Both bugs existed in undo.js (changeSelectedAttributeNoUndoMethod) and history.js (ChangeElementCommand.apply/unapply). The history.js version was worse — it replaced the entire transform attribute with just rotate(...), destroying any other transforms.

## To replicate the bug:
Use this sample SVG: compound-transform-bug

  This branch is deployed to: https://geoffyoungs.github.io/svgedit/
  Compare with the upstream demo at https://svgedit.netlify.app/editor/index.html

  To test the bug:
  1. Open both editors
  2. Import the pinwheel SVG (or draw shapes with compound transforms)
  3. Select the group and change stroke-width
  4. On the upstream demo, the shapes jump — on the fixed version, they stay in place

## Checklist

NB: Claude Code was used to identify and fix this issue.

## Summary by Sourcery

Fix rotation center recalculation so that transforms on rotated elements are preserved and compound transforms remain stable when attributes change or history operations are applied.

Bug Fixes:
- Restrict rotation center recalculation to geometry-affecting attributes and compute the new center using only post-rotation transforms, preventing corrupt rotation on elements with compound transforms.
- Preserve existing transform lists when updating rotation centers during undo/redo instead of replacing the entire transform with a simple rotate(), avoiding loss of translate/scale transforms.

Tests:
- Add visual fixtures and a Playwright-based screenshot script to document and manually verify the rotation recalculation behaviour on elements with compound transforms.


jfhenon:
plz fix the linter errors so I can merge your (very useful) PR

geoffyoungs:
Apologies - I should have checked that before submitting.  SVG-Edit is an excellent and mature tool - I was both surprised and excited to find a legitimate bug!

jfhenon (approved):

PR #1091, opened 2026-07-11
Fix move/resize/undo geometry flattening (#1090, #1086) + dependency updates

### LINKED ISSUE #1086: *breaking* shape x y coordinates, width / height inputs do not update and shape resizing does not work in accordance with SVG
Hello,

The issue i am describing today is regarding the the following three problems:

1. **Shape coordinates**
Creating any shape, be it rectangle, path etc. and moving it around does not update the inputs for the coordinates.

2. **Shape sizing**
Resizing the shape does not update the width and height inputs (for shapes that support this).

3. **Shape scaling/ stretching**
Resizing the shapes should recalculate the shape itself (like the previous versions of the editor did). Currently, they are being scaled. This means that the proportions of the shapes are not kept. This negates the purpose of SVG.

For the example, I have provided two screenshots with a simple rectangle to demonstrate. 
**Figure 1** is the before and **Figure 2** is the after for the example.

**Figure 1**
Image

**Figure 2**
Image

Regards.

### LINKED ISSUE #1090: Undo command on resizing shapes
**PLEASE NOTE: We are trying to make this project live again. While you can file an issue, we will do our best but you should not expect any action, even if we label the issue. However, we are welcoming new contributors **

**Describe the bug**
When resizing shapes or rotating them, when clicking undo or using ctrl-z, the shape does not revert to the original size in the location it was moved to. It instead jumps back to the position it was in when placed or last in before being moved and then resized. I also noticed this only seems to occur on the ellipse and star tools, the rectangle seems unaffected. 
It is like some of the shapes states are not properly recorded.

**To Reproduce**
1. Go to main canvas
2. Click on ellipse tool or star tool
3. Draw a shape
4. Move the shape to a new location
5. Resize the shape
6. Undo the resize action
7. See Error (the shape returns to its first location)

**Expected behavior**
I expected that after using ctrl-z to undo the resize, the shape would return to the size it was before and remain in the same spot I placed it.

**SVG-Edit environment (IMPORTANT)**
- File for SVG-Edit: N/A
- Version: 7.3.3
- Protocol: ??

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Firefox
 - Version: 152.0.5

**Screenshots**

**Log**
If applicable, copy any error logs in your browser console. This can be very
helpful in identifying the problem.

**Additional context**

jfhenon:
## Summary

- Fixes #1090: undoing a resize that followed a move reverted the shape all the way back to its creation-time position instead of just the resize.
- Fixes #1086: resizing a shape never updated its real geometry attributes (`x`/`y`/`width`/`height`/`rx`/`ry`/etc.), so the coordinate/size panel inputs never reflected a resize and shapes appeared scaled (distorting stroke width) instead of recalculated.
- Updates all workspace dependencies to latest, including majors (vite 7→8, jsdom 27→29, i18next 25→26, react/react-dom 19.1→19.2, nyc 17→18, open-cli 8→9, vite-plugin-istanbul 7→9, `@babel/preset-react` 7→8).

## Root cause (#1090 / #1086)

Both issues stem from the same defect in `packages/svgcanvas/core/event.js`'s mouseup handler:

1. The dummy transform inserted at the start of a move-drag was an identity **matrix**, but the code that replaces it during dragging only recognized dummies of type **translate** — leaving a stray extra transform behind after every move and routing simple moves into a "consolidate transforms" path instead of flattening them into real geometry.
2. That consolidate path (shared by move/resize/rotate) always treated a transform list with more than one item as a "pre-existing multi-transform" case to collapse into one matrix — but a resize *always* builds an intentional translate/scale/translate list of exactly that shape. So `recalculateDimensions()` (which already has a dedicated branch for this exact shape) was never actually called for a resize, and the "old transform" recorded for undo was read from a map that's only populated during move drags, so undoing a resize always recorded `''` as the prior state instead of the move's real transform.

**Fix:** the drag-start dummy is now a real `translate(0,0)` so it's correctly recognized and replaced; and non-group elements now always try `recalculateDimensions()` first (falling back to raw matrix consolidation only for groups or shapes it doesn't recognize), using the transform captured at mousedown as the true "before" state for undo regardless of interaction mode.

## Test plan

- [x] `npm run lint`
- [x] `npx vitest run tests/unit` — 562 tests pass, including new regression test `tests/unit/move-resize-undo.test.js` covering move→resize→undo and resize-only flattening for `rect` and `ellipse`
- [x] `npm run build` — production build succeeds

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Fix move/resize/undo handling so drag transforms are correctly flattened into element geometry, ensuring accurate undo behavior and geometry attributes, and update key workspace dependencies.

Bug Fixes:
- Ensure move followed by resize preserves the moved position when undoing the resize instead of reverting to the original creation-time geometry.
- Ensure resize operations update elements’ real geometry attributes and do not leave scale transforms that visually distort stroke width.
- Ensure group elements consolidate drag transforms into a single matrix while preserving intended behavior for non-group elements.

Enhancements:
- Refine transform handling to prioritize geometry flattening via recalculateDimensions for non-group elements, with a matrix consolidation fallback.
- Improve jsdom-based test reliability by shimming missing SVG APIs used during mouse interactions.

Build:
- Update core build and tooling dependencies (vite, jsdom, nyc, open-cli, vite-plugin-istanbul, rollup optional binary) to current versions.

Tests:
- Add unit regression tests covering move→resize→undo and resize-only flows for rect, ellipse, path, and group elements.
- Adjust Playwright layer panel e2e expectations to align with the latest @playwright/test behavior.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,90 +16,79 @@
         "@svgedit/svgcanvas": "7.4.1",
         "browser-fs-access": "0.38.0",
         "elix": "15.0.1",
-        "i18next": "25.7.4",
-        "jspdf": "4.0.0",
+        "i18next": "26.3.6",
+        "jspdf": "4.2.1",
         "pathseg": "1.2.1",
         "svg2pdf.js": "2.7.0"
       },
       "devDependencies": {
-        "@playwright/test": "^1.57.0",
+        "@playwright/test": "^1.61.1",
         "@rollup/plugin-dynamic-import-vars": "2.1.5",
-        "@vitest/coverage-v8": "^4.0.16",
+        "@vitest/coverage-v8": "^4.1.10",
         "jamilih": "0.63.1",
         "jsdoc": "4.0.5",
-        "jsdom": "^27.4.0",
+        "jsdom": "^29.1.1",
         "npm-run-all": "4.1.5",
-        "nyc": "17.1.0",
-        "open-cli": "8.0.0",
+        "nyc": "18.0.0",
+        "open-cli": "9.0.0",
         "remark-cli": "12.0.1",
         "remark-lint-ordered-list-marker-value": "4.0.1",
-        "rimraf": "6.1.2",
+        "rimraf": "6.1.3",
         "standard": "17.1.2",
-        "vite": "^7.3.1",
-        "vite-plugin-istanbul": "^7.2.1",
+        "vite": "^8.1.4",
+        "vite-plugin-istanbul": "^9.0.1",
         "vite-plugin-string": "^1.2.3",
-        "vitest": "^4.0.16"
+        "vitest": "^4.1.10"
       },
       "engines": {
         "node": ">=20"
       },
       "optionalDependencies": {
-        "@rollup/rollup-linux-x64-gnu": "4.55.1"
+        "@rollup/rollup-linux-x64-gnu": "4.62.2"
       }
     },
-    "node_modules/@acemir/cssom": {
-      "version": "0.9.30",
-      "resolved": "https://registry.npmjs.org/@acemir/cssom/-/cssom-0.9.30.tgz",
-      "integrity": "sha512-9CnlMCI0LmCIq0olalQqdWrJHPzm0/tw3gzOA9zJSgvFX7Xau3D24mAGa4BtwxwY69nsuJW6kQqqCzf/mEcQgg==",
-      "dev": true,
-      "license": "MIT"
-    },
     "node_modules/@asamuzakjp/css-color": {
-      "version": "4.1.1",
-      "resolved": "https://registry.npmjs.org/@asamuzakjp/css-color/-/css-color-4.1.1.tgz",
-      "integrity": "sha512-B0Hv6G3gWGMn0xKJ0txEi/jM5iFpT3MfDxmhZFb4W047GvytCf1DHQ1D69W3zHI4yWe2aTZAA0JnbMZ7Xc8DuQ==",
+      "version": "5.1.11",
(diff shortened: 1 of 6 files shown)


jfhenon:
Addressed the Sourcery review feedback:

- **jsdom `replaceItem` stub**: reimplemented in terms of the list's own `removeItem`/`insertItemBefore` public API instead of mutating the internal `_items` array directly.
- **Resize-only undo coverage**: added an undo assertion to the resize-only test case verifying it restores the original geometry with no leftover transform.
- **Broader element/interaction coverage**: added `path` alongside `rect`/`ellipse` for the move→resize→undo and resize-only flows, plus a new test covering the `isGroup` consolidation branch (which `recalculateDimensions()` intentionally skips for groups).
- **npm audit**: ran `npm audit fix`, resolving all 14 flagged vulnerabilities in transitive dev-tooling deps (eslint/remark/nyc chains) - 0 vulnerabilities remain, no direct dependency versions changed.
- **BlueOak-1.0.0 license flags** (`minipass`, `spawn-wrap`): checked - this is their genuine, current license at every supported version (not stale metadata), a legitimate permissive OSS license that some scanners simply don't have indexed yet. Downgrading to reintroduce a "recognized" license would mean reintroducing the vulnerabilities just fixed above, so left as-is.
- **Splitting into separate PRs**: per explicit request this round combines the #1090/#1086 fix with the dependency updates; keeping as one PR.

All changes verified with `npm run lint`, `npm test` (unit + e2e), and `npm run build`.

jfhenon:
@sourcery-ai review

jfhenon:
@sourcery-ai review

PR #1098, opened 2026-07-15
Improve SVG import and serialization for attributes, text, and reused elements

veselin-kutsarov:
Do not iterate attributes in reverse order!
Properly parse attribute value if it is a number!
Do not skip data attribute even they are empty!
Do not trim values of <text> tags.
Allow not  to use existing imported elements.
Extend SVG whitelist.

## Checklist

## Summary by Sourcery

Improve SVG import/export behavior for attributes, text nodes, and reuse of previously imported elements.

New Features:
- Add a flag to SVG import to optionally avoid reusing previously imported elements.
- Extend the SVG sanitization whitelist to support alignment-baseline on text and tspan elements.

Bug Fixes:
- Preserve attribute iteration order and correctly handle numeric attribute values during SVG serialization.
- Retain empty data-* attributes in serialized SVG output instead of dropping them.
- Avoid trimming whitespace inside <text> nodes during SVG serialization and sanitization to preserve text content formatting.

Enhancements:
- Align gradient conversion logic and related queries without changing behavior.


jfhenon (approved):

jfhenon:
We need to adapt the test that checks for trimming values.

jfhenon:
maybe better to adapt the test and to make sure the string is unchanged (so future change to the code will not impact your change)

veselin-kutsarov:
> * For numeric attribute handling, you are checking `!isNaN(attrVal) && !isNaN(parseFloat(attrVal))` but still pass the original string to `shortFloat`; it may be clearer and more robust to call `shortFloat` on a parsed number and ensure mixed units like `"10px"` don’t get incorrectly treated as bare numbers.
> 
shortFloat already converts numeric strings with Number(val) internally, so passing the original numeric string is intentional and equivalent to passing a parsed number. Mixed-unit values such as "10px" are not accepted by this branch because isNaN("10px") is true. The additional parseFloat check prevents empty or whitespace-only strings from being treated as zero.

jfhenon (approved):

PR #1099, opened 2026-07-16
Improve SVG transform handling and numeric toolbar input behavior

veselin-kutsarov:
## Summary

This PR improves SVG transform handling across editing workflows and makes numeric toolbar inputs safer and more predictable.

### Transform handling

- Normalize transforms after drag, resize, rotation, programmatic movement, and ungrouping.
- Recalculate rotation centers after geometry changes to prevent elements from jumping when rotation starts.
- Bake representable transforms into shape geometry while preserving residual matrices required for skew/shear.
- Keep groups, text, and `<use>` elements in a normalized `rotate(...) matrix(...)` form when necessary.
- Normalize transforms pushed to child elements during ungrouping without changing text `font-size`.
- Avoid creating identity transforms on clicks or zero-delta pointer movements.
- Preserve existing leading translations when dragging imported groups.
- Correct transformed bounding-box calculations used by area selection.
- Emit final `changed` and `selected` events so toolbar values and multi-selection actions remain synchronized.
- Avoid rewriting rotation transforms when the requested angle has not changed.

### Numeric spin inputs

- Select the current value when an input receives focus.
- Keep manually typed values local until Enter or a valid blur.
- Restore the previous value on Escape or invalid input.
- Preserve immediate updates from increment/decrement buttons.

## Verification

- Full unit suite: 45 test files, 605 tests passed.
- Playwright end-to-end suite: 81 tests passed.
- Lint passed.
- Production build passed.
- Added focused coverage for transforms, rotation centers, ungrouping,
  selection events, transformed bounding boxes, undo history, drag thresholds,
  and spin-input behavior.

## Checklist

- [x] Added Playwright UI tests.
- [x] Ran `npm test`; lint, unit tests, coverage collection, production build,
      and Playwright tests passed.
- [ ] Added user documentation. Not applicable for this PR because it changes
      internal editing behavior and is documented through focused unit and
      Playwright regression tests.

## Summary by Sourcery

Improve SVG transform normalization across editing workflows and make numeric spin inputs commit changes more safely and predictably.

New Features:
- Normalize and preserve group, text, and use element transforms into stable rotate-plus-matrix forms while baking affine transforms into geometry where possible.
- Introduce rotation math utilities for normalizing angles and extracting rotation centers to support consistent transform handling.
- Add controlled commit/revert behavior to the SESpinInput component so manual edits are only applied on explicit confirmation.

Bug Fixes:
- Prevent creation of stray identity drag transforms or unnecessary matrices during small pointer movements and clicks.
- Ensure recalculateDimensions returns concrete undoable commands for groups, text, use elements, and shapes instead of silently bailing out.
- Fix transformed bounding-box calculations so pure translations and identity transforms are handled correctly for selection and clipping.
- Preserve text font-size and referenced positioning when normalizing transforms during ungrouping and recalc flows.
- Ensure drag, move, resize, and multiselect mouse-up flows emit final changed/selected events only when there were actual geometry or transform updates.

Enhancements:
- Refine recalculateDimensions to bake transforms into geometry when safe, preserve off-axis residual matrices, and recenter rotation transforms after drag/resize or programmatic moves.
- Improve group, text, and use transform handling so rotation centers and visual placement remain stable across drag, resize, flip, ungroup, and toolbar rotation operations.
- Normalize pushed-down transforms during ungroup so children receive well-structured rotate-plus-matrix transform lists suitable for further editing.
- Update toolbar rotation handling to avoid no-op transform rewrites, maintain visual rotation centers, and capture baked-geometry changes as single undoable steps.
- Harden Playwright/NYC E2E tooling and coverage seeding for cross-platform execution and cleaner coverage output.

Tests:
- Add extensive unit tests for recalculateDimensions covering rotated/scaled shapes, groups, text, use, clip-paths, and off-axis residual transforms.
- Add unit tests for event drag thresholds, transform normalization, and final changed/selected event emission after selection interactions.
- Introduce tests for programmatic group movement and ungrouping to verify transform normalization and text font-size preservation.
- Add math utility tests for rotation angle normalization, rotation center recovery, and rotation transform summarization.
- Add unit and E2E tests validating toolbar rotation behavior, including undo/redo of baked geometry changes.
- Add focused unit tests for the SESpinInput component to cover manual entry, commit/revert behavior, and interaction with spin buttons.
- Extend bounding-box tests to cover transformed and translated elements and verify visual bounds preservation during recalc.


jfhenon (approved):

jfhenon:
Thanks for the detailed PR — the transform-normalization work and the spin-input rewrite are both solid, and the added unit/e2e coverage is appreciated. A few things worth clarifying/addressing before merge:

## 1. Drag threshold: 4px → 1px — please confirm intent

In `event.js` (`deltaThreshold`), the threshold was lowered from `4` to `1`, with the comment changed from "moved at least 4 px" to "a deliberate pointer movement." This isn't mentioned in the PR description, which only talks about avoiding stray identity transforms on *zero*-delta movement/clicks — a 1px threshold is a much bigger behavioral change than that goal requires.

Was lowering the threshold to 1px intentional, or should it have stayed at 4px (or somewhere in between) while only the "don't insert a drag transform until the threshold is reached" logic changed? At 1px, trackpad jitter or a slightly unsteady click is enough to register as a drag — exactly the accidental-move problem the original 4px threshold was meant to prevent. If this was just to avoid an off-by-one in a test, it'd be better to keep the threshold at 4 and fix the test instead.

## 2. Make the shape/path baking branch consistent about no-op detection

The group branch and the text/tspan/use branch of `recalculateDimensions` both snapshot `transformListSignature(tlist)` before normalizing, normalize, and only build a `ChangeElementCommand` / return a batch command if the signature actually changed — otherwise they return `null` (no history entry).

The generic shape/path branch doesn't do this: `bakeTransformsIntoAttributes` returns `true` whenever `tlist` was non-empty going in, and the caller unconditionally pushes a `ChangeElementCommand` whenever that's `true` — even if baking happens to leave the element in an equivalent state.

Suggestion: capture a "did anything actually change" snapshot (transform signature plus the geometry attrs baking can touch — `x/y/width/height/d`, etc.) before calling `bakeTransformsIntoAttributes`, and only push the undo command if something differs afterward, matching what the group and text/use branches already do. This avoids polluting the undo stack with no-op entries for plain shapes.

## 3. Consolidate the `getStartTransform`/`setStartTransform` guard-and-restore pattern

This shape appears three times (once in `event.js`'s `mouseUpEvent`, twice in `selected-elem.js`'s `moveSelectedElements`/`normalizePushedGroupChild`/`flipSelectedElements`):

```js
const hasStartTransformApi = typeof svgCanvas.getStartTransform === 'function' &&
  typeof svgCanvas.setStartTransform === 'function'
const saved = hasStartTransformApi ? svgCanvas.getStartTransform() : null
if (hasStartTransformApi) svgCanvas.setStartTransform(oldTransform)
try {
  cmd = svgCanvas.recalculateDimensions(elem)
} finally {
  if (hasStartTransformApi) svgCanvas.setStartTransform(saved)
}
```

The `typeof ... === 'function'` guard only exists to tolerate test doubles for `svgCanvas` that don't implement `getStartTransform`/`setStartTransform` — the real canvas always has them.

Two suggestions:
- Extract the save/set/call/restore sequence into one shared helper, e.g.:
  ```js
  export const withStartTransform = (transform, fn) => {
    const saved = svgCanvas.getStartTransform()
    svgCanvas.setStartTransform(transform)
    try {
      return fn()
    } finally {
      svgCanvas.setStartTransform(saved)
    }
  }
  ```
  and call `withStartTransform(oldTransform, () => svgCanvas.recalculateDimensions(elem))` at all three sites.
- Drop the `typeof ... === 'function'` runtime guards, and instead update the test mocks/stubs to implement `getStartTransform`/`setStartTransform` (even as trivial no-ops). Production code shouldn't need to feature-detect its own internal API to accommodate incomplete test doubles.

jfhenon (approved):

